PR #8, opened 2017-12-21
MM task from 51 to 60

hsp-mona:
MM task from 51 to 60


src/main/java/am/hsp/apps/seua/Task_58.java
```diff
@@ -0,0 +1,102 @@
...
+public class Task_58 {
+	private static final int SUCCESS_CASE = 5;
+
+	public static void main(String[] args) {
+		char f;
+
+		int number = getUserInput("Please enter the number. ");
+
+		if (isThreeDigit(number)) {
+			if (isGreaterThanSuccessCase(number)) {
```
hsp-harut: You can simplify this if/else by using **? :**

src/main/java/am/hsp/apps/seua/Task_58.java
```diff
@@ -0,0 +1,102 @@
...
+		} else {
+			System.out.println("Your input number isn't three-digit!");
+		}
+	}
+
+	/**
+	 * Checks is the sum of hundreds and tens of number greater than success case or
+	 * not
+	 */
+	private static boolean isGreaterThanSuccessCase(int number) {
```
hsp-harut: It would be great if method name will say what is actually greater than success case!

src/main/java/am/hsp/apps/seua/Task_59.java
```diff
@@ -0,0 +1,150 @@
...
+ * Print the digits of input number in ascending order.
+ * 
+ * @author Mona
+ */
+public class Task_59 {
+	public static void main(String[] args) {
+		int number = getUserInput("Please enter the number. ");
+		
+		if (isThreeDigit(number)) {
+			System.out.print(getMinimumDigit(number) + " ");
```
hsp-harut: You are calculating digits in each of these methods, which is actually resource consuming process!
So it would be correct when you will calculate digits only once and keep them in variables, and then you can just use those variables for calculating min/mid/max...

src/main/java/am/hsp/apps/seua/Task_60.java
```diff
@@ -0,0 +1,151 @@
...
+ * 
+ * @author Mona
+ */
+
+public class Task_60 {
+	public static void main(String[] args) {
+		int number = getUserInput("Please enter the number. ");
+
+		if (isThreeDigit(number)) {
+			System.out.print(getMaximumDigit(number) + " ");
```
hsp-harut: Same issues as in previous task!

PR #11, opened 2017-12-22
DP tasks 21 to 30

hsp-davit:


src/main/java/am/hsp/apps/seua/Task_26.java
```diff
@@ -0,0 +1,63 @@
...
+	}
+	
+	/**
+	 * checks  the  has even number or not.
+	 * @param number1
+	 * @param number2
+	 * @param number3
+	 * @return
+	 */
+	private static boolean isEvenOrOdd (int number1, int number2, int number3) {
```
hsp-harut: Based on your method name it seems that it should return true for both cases - even and odd.
But actually your method is for checking the number is Even or NOT (when false - that means that it is an Odd number).
So it is better to change your method name, and indicate that it needs to check are any of the numbers is Even or not

src/main/java/am/hsp/apps/seua/Task_26.java
```diff
@@ -0,0 +1,64 @@
...
+
+	public static void main(String[] args) {
+
+		int number1 = getUserInput("please enter the number1");
+		int number2 = getUserInput("please enter the number2");
+		int number3 = getUserInput("please enter the number3");
+
+		int result = 2;
+
+		if (isAnyEven(number1, number2, number3)) {
```
hsp-harut: Try to make this even simpler by using "? :" instead of if/else

PR #29, opened 2018-01-17
Sat... 161-170

hsp-satenik:
Implemented from Task_161 to Task_170.


src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -93,21 +100,127 @@ private int getMin(int number1, int number2) {
...
-		
+
 		return number1;
 	}
-	
-	
+
 	protected int pow(int number, int power) {
-		if(power == 0) {
+		int result = 1;
```
hsp-harut: define variables where you actually need them

PR #33, opened 2018-01-19
Mona: Tasks from 201 to 210

hsp-mona:
Mona: Tasks from 201 to 210


src/main/java/am/hsp/apps/seua/Task_208.java
```diff
@@ -0,0 +1,48 @@
...
+		System.out.print("Instance that digits of input number are equal to each other ");
+		System.out.println(isNDigit(number, 1) ? "not correct" : areDigitsEqual(number));
+	}
+
+	/**
+	 * Check are digits of input number equal to each other or not.
+	 */
+	private boolean areDigitsEqual(int number) {
+		int count = countOfDigit(number);
+		while (isDigitEqualToPrevious(number, count)) {
```
hsp-harut: Again, similar issue as in previous task. 
Try to simplify this. I would suggest to keep the last digit of the number, and then write a while loop to take the last digit of the number in each iteration and compare it with the very first last digit we have extracted before the loop. If they are not equal - just return false.

src/main/java/am/hsp/apps/seua/Task_209.java
```diff
@@ -0,0 +1,39 @@
...
+
+		System.out.println("Instance that any digit of input number is odd is " + isAnyDigitOdd(number));
+	}
+
+	/**
+	 * Checks is any digit of input number odd or not.
+	 */
+	private boolean isAnyDigitOdd(int number) {
+		int count = countOfDigit(number);
+		while (isEven(getDigitByIndex(number, count))) {
```
hsp-harut: Same issue as in previous tasks... There is no need to calculate count, and also there is no need to get digits by any specified indexes. You just need to get last digit per iteration and return true if it is odd...

src/main/java/am/hsp/apps/seua/Task_210.java
```diff
@@ -0,0 +1,82 @@
...
+	private boolean isSumOddIndexEqualSumEvenIndexDigits(int number) {
+		
+		return (sumOfOddIndexDigits(number) == sumOfEvenIndexDigits(number));
+	}
+	
+	
+	/**
+	 * get the sum of digits by even index.
+	 */
+	private int sumOfEvenIndexDigits(int number) {
```
hsp-harut: 1. First of all this is DRY violation with the below method. The only difference between these 2 methods is only the starting value for "i" variable. So even if you will go by this approach for solving this task, you need to create only one method lets say getSumOfDigitsFromIndex(int number, int startIndex) and use this method inside sumOfEvenIndexDigits and sumOfOddIndexDigits methods...

2. I guess you can simplify overall solution for this task. You just need to have one "while" loop and get each time the last digit. You need to store current index of iteration, and also 2 sum variables - one for odd sum and the secind for even sum. In each iteration you need to check whether currentIndex is even or odd and add the digit in appropriate sum variable. After the while loop just return sumOdd == sumEven.

src/main/java/am/hsp/apps/seua/Task_208.java
```diff
@@ -0,0 +1,39 @@
...
+		System.out.print("Instance that digits of input number are equal to each other ");
+		System.out.println(isNDigit(number, 1) ? "not correct" : areDigitsEqual(number));
+	}
+
+	/**
+	 * Check are digits of input number equal to each other or not.
+	 */
+	private boolean areDigitsEqual(int number) {
+		while (number != 0) {
+			if (getLastDigit(number) != getLastDigit(number / 10)) {
```
hsp-harut: I guess this method will always return false... try to understand why...
ALWAYS test your code before pushing it into Git.

src/main/java/am/hsp/apps/seua/Task_210.java
```diff
@@ -0,0 +1,64 @@
...
+	 
+	private boolean isSumOddIndexEqualSumEvenIndexDigits(int number) {
+
+		return (sumOfDigitsFromIndex(number, 1) == sumOfDigitsFromIndex(number, 2));
+	}
+
+	/**
+	 * get the sum of digits by odd index.
+	
+	private int sumOfDigitsFromIndex(int number, int startIndex) {
```
hsp-harut: Based on method name it is not clear that this will iterate through the digits by 2 step. It would be better to
have a method name like "sumOfEvenOrOddDigits(int number, boolean evenDigits)" - and pass a boolean instead of start index, then in the method you will initialize startIndex based on boolean flag.

src/main/java/am/hsp/apps/seua/Task_208.java
hsp-mona: I test but everything is ok. for 8888 is print true. for 7856 is print false.

hsp-mona: i understand. test is ok cause in 14line Task_210 task = new Task_210();.

PR #54, opened 2018-01-30
Davit: tasks 211-220

hsp-davit:


src/main/java/am/hsp/apps/seua/Task_215.java
```diff
@@ -0,0 +1,42 @@
...
+		System.out.println(getSumElementsOfEvenIndex(array));
+	}
+
+	/**
+	 * get the sum of the elements of the even index
+	 * 
+	 * @param array
+	 * @return
+	 */
+	protected double getSumElementsOfEvenIndex(int[] array) {
```
hsp-harut: Apri Davon :+1:

src/main/java/am/hsp/apps/seua/Task_215.java
```diff
@@ -0,0 +1,42 @@
...
+		int sum = 0;
+		for (int i = 0; i < array.length; i += 2) {
+			sum += (array[i]);
```
hsp-harut: braces not required when actually there is only one statement

src/main/java/am/hsp/apps/seua/Task_218.java
```diff
@@ -0,0 +1,45 @@
...
+	 * get the sum of absolute values of odd index
+	 * 
+	 * @param array
+	 * @return
+	 */
+	protected double getSumOfAbsoluteValuesOfOddIndex(int[] array) {
+		int sum = 0;
+		for (int i = 1; i < array.length; i += 2) {
+			if (array[i] < 0) {
+				array[i] *= -1;
```
hsp-harut: You are NOT eligible  to update actual array!!!
You just need to calculate sum of absolute values.

You need to write a method for returning absolute value ov given number - and use that method here...

src/main/java/am/hsp/apps/seua/Task_220.java
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	@Override
+	public void solve() {
+		int length = getUserInput("Please enter the length of array:");
+		int[] array = getIntArray(length);
+
+		printArray(array);
+
+		System.out.println(getCountPositiveElements(array));
```
hsp-harut: You should inform user which number corresponds to which - so Add also here a message lets say "count of negative elements:"...

PR #55, opened 2018-01-30
Davit: tasks 221-230

hsp-davit:


src/main/java/am/hsp/apps/seua/TaskArray.java
```diff
@@ -0,0 +1,72 @@
...
+		System.out.println();
+	}
+
+	/**
+	 * get absolute value
+	 * 
+	 * @param number
+	 * @return
+	 */
+	protected int getAbsoluteValue(int number) {
```
hsp-harut: Actually this method is more generic - and not related to especially Arrays

src/main/java/am/hsp/apps/seua/TaskArray.java
```diff
@@ -0,0 +1,72 @@
...
+
+	/**
+	 * get border of numbers
+	 * 
+	 * @param number
+	 * @param minimum
+	 * @param maximum
+	 * @return
+	 */
+	protected boolean getBorderOfNumbers(int number, int minimum, int maximum) {
```
hsp-harut: same as for above method.
Also bad method name - when it starts with "get" - it means that it need to return some data - it is not clear that it needs to return a boolean.
Your method name should indicate some condition...

src/main/java/am/hsp/apps/seua/Task_226.java
```diff
@@ -0,0 +1,46 @@
+package main.java.am.hsp.apps.seua;
+
+/**
+ * get count of elements which absolute value is smaller than input value.
+ * 
+ * @author davit
+ *
+ */
+public class Task_226 extends TaskArray {
```
hsp-harut: Actually very similar to the previous task. For future please skip such tasks when there is nothing for you to learn.

src/main/java/am/hsp/apps/seua/Task_229.java
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	@Override
+	public void solve() {
+		int length = getUserInput("Please enter the length of array:");
+		int[] array = getIntArray(length);
+
+		printArray(array);
+
+		System.out.println(getProductWhenDifferencePositive(array));
```
hsp-harut: Always add some informative message to user before displaying the result:
System.out.println("Product of elements with value greater than their index in the array is equal to: " + getProductWhenDifferencePositive(array));

PR #62, opened 2018-02-01
Gevor: tasks 201-210

hsp-gevorg:


src/main/java/am/hsp/apps/seua/Task_206.java
```diff
@@ -0,0 +1,46 @@
...
+	 * get number right than left
+	 * 
+	 * @param number
+	 * @return
+	 */
+	public int getNumberRightThanLeft(int number) {
+		int result = 0;
+		int count = getDigitCount(number);
+		int sum = 0;
+		for (int i = 1; i <= getDigitCount(number); i++) {
```
hsp-harut: Actually you can get last digit of the number much easier rather than calculate digit count, and then in each iteration extract required last digit. This is resource consuming. You must use the same approach for extracting last digit in each iteration as you did in Task 204. 
Please try to explain why you implemented this task this way - when you had simpler and more correct implementation of similar functionality in Task 204?

PR #69, opened 2018-02-14
Mona: Searching Algorithms

hsp-mona:
Mona: Implementation Of Searching Algorithms


src/main/java/am/hsp/apps/seua/search/InterpolationSearchByRecursion.java
```diff
@@ -0,0 +1,55 @@
...
+
+	public int getSearchingNumberIndex(int[] array, int number) {
+
+		return interpolationSearch(array, number, 0, array.length - 1);
+	}
+
+	public int interpolationSearch(int[] array, int number, int start, int end) {
+		if (array[start] == number) {
+			return start;
+		} else if (start == end) {
```
hsp-harut: why you need "else" ?

src/main/java/am/hsp/apps/seua/search/InterpolationSearchByRecursion.java
```diff
@@ -0,0 +1,55 @@
...
+		task.solve();
+	}
+
+	@Override
+	public void solve() {
+		int[] array = { 1, 2, 4, 6, 7, 10, 11, 14, 15 };
+		printGivenArray(array);
+		int number = getUserInput("Enter element to search for :");
+
+		System.out.println((getSearchingNumberIndex(array, number) == -1) ? "The element is not found"
```
hsp-harut: same

src/main/java/am/hsp/apps/seua/search/InterpolationSearch.java
```diff
@@ -0,0 +1,55 @@
...
+		task.solve();
+	}
+
+	@Override
+	public void solve() {
+		int[] array = { 1, 2, 4, 6, 7, 10, 11, 14, 15 };
+		printGivenArray(array);
+		int number = getUserInput("Enter element to search for :");
+
+		System.out.println((getSearchingNumberIndex(array, number) == -1) ? "The element is not found"
```
hsp-harut: same

src/main/java/am/hsp/apps/seua/search/BinarySearch.java
```diff
@@ -0,0 +1,58 @@
...
+
+	private int doBinarySearch(int[] array, int left, int right, int number) {
+		if (left <= right) {
+			int middle = (left + right) / 2;
+
+			if (array[middle] == number) {
+				return middle;
+			}
+
+			if (array[middle] < number) {
```
hsp-harut: Please pay attention on my previous comment I wrote related this. You may think that this is not important - but actually no. You must write the code the way you actually think.

"_it would be easier to understand if you will write the same as "number > array[middle]",
because when you think about the algorithm, you are thinking this way "if my number is greater than the middle element, then it should be on the right side from middle". So the first part you think is related to "number" and not to "middle" (if my number is greater than middle...). So this means that your code also needs to be according to the way you think: So "if my number is greater than middle" == number > array[middle], and not vice versa._"

PR #71, opened 2018-02-15
Gevor: tasks 211- 220

hsp-gevorg:


src/main/java/am/hsp/apps/seua/Task_217.java
```diff
@@ -0,0 +1,40 @@
+package main.java.am.hsp.apps.seua;
+
+import main.java.am.hsp.apps.seua.base.TaskArray;
+
+/**
+ * get sum of odd index pow
```
hsp-harut: Product of array elements which have odd index.

src/main/java/am/hsp/apps/seua/Task_218.java
```diff
@@ -0,0 +1,45 @@
+package main.java.am.hsp.apps.seua;
+
+import main.java.am.hsp.apps.seua.base.TaskArray;
+
+/**
+ * get sum of odd total index
```
hsp-harut: Get sum of absolute values of elements with odd indexes

src/main/java/am/hsp/apps/seua/Task_218.java
```diff
@@ -0,0 +1,45 @@
...
+	 * get sum of odd total index
+	 * 
+	 * @param array
+	 * @return
+	 */
+	public int getSumOfOddTotalIndex(int array[]) {
+		int sum = 0;
+		for (int i = 1; i < array.length; i += 2) {
+			if (array[i] < 0) {
+				array[i] *= (-1);
```
hsp-harut: it is better to have a method which will calculate the absolute value of integer - similar to the method that is available in Math library "int abs(int)"

PR #72, opened 2018-02-16
Gevor: tasks 221- 230

hsp-gevorg:


src/main/java/am/hsp/apps/seua/Task_221.java
```diff
@@ -0,0 +1,46 @@
+package main.java.am.hsp.apps.seua;
+
+import main.java.am.hsp.apps.seua.base.TaskArray;
+
+/**
+ * get sum of all numbers middle given numbers
```
hsp-harut: between given range (Not middle given numbers)

src/main/java/am/hsp/apps/seua/Task_221.java
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * get sum of all numbers middle given numbers
+	 * 
+	 * @param array
+	 * @param from
+	 * @param to
+	 * @return
+	 */
+	public int getSumOfMiddleGivenNumbers(int array[], int from, int to) {
```
hsp-harut: getSumNumbersForGivenRange

src/main/java/am/hsp/apps/seua/Task_224.java
```diff
@@ -0,0 +1,47 @@
+package main.java.am.hsp.apps.seua;
+
+import main.java.am.hsp.apps.seua.base.TaskArray;
+
+/**
+ * get numbers cube sum small than given number
```
hsp-harut: Get sum of number cubes when absolute value of the number is less than given number

src/main/java/am/hsp/apps/seua/base/TaskArray.java
```diff
@@ -19,11 +19,26 @@
...
+	/**
+	 * get double array
+	 * 
+	 * @param length
+	 * @return
+	 */
+	protected double[] getDoubleArray(int length) {
+		double array[] = new double[length];
+		for (int i = 0; i < length; ++i) {
 			array[i] = getRandomInt(FROM, TO);
```
hsp-harut: You will have here anint array not the double array - as you getRandomInt method ALWAYS return an Integer and not a double. You need to create another method for getting random Double and use that method here...

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -213,6 +213,23 @@ protected int pow(int number, int degree) {
...
 		return result;
 	}
+	
+	/**
+	 * The number degree
+	 * 
+	 * @param number
+	 * @return
+	 */
+	protected double pow(double number, int degree) {
```
hsp-harut: DRY violation with the above method.
You can use this method in the method above to not violate DRY principle... Or you can just remove old method,
as you can use this one for the int number as well.

PR #79, opened 2018-02-27
Vahagn_Task 171 178 recurcion

hsp-vahagn:
Task 171 178 recurcion


src/main/java/am/hsp/apps/seua/Task_176_recursion.java
```diff
@@ -0,0 +1,49 @@
...
+		int number = getUserInput("Please neter the number");
+		System.out.println("Print Fibonachi numbers");
+		printSequenceValuesOfFibonachi(number);
+	}
+
+	/**
+	 * get Fibonachi numbers
+	 */
+	private int getFibonachiNumbers(int number) {
+		if (number == 1) {
```
hsp-harut: Can't you make these 2 'if' statements simpler?

src/main/java/am/hsp/apps/seua/Task_174_recursion.java
```diff
@@ -0,0 +1,49 @@
+package main.java.am.hsp.apps.seua;
+
+import main.java.am.hsp.apps.seua.base.TaskBase;
+
+/**
+ * print sequence x1,x2,...,xN where Xk = 2 + 1 / X(k - 1) of real numbers by
```
hsp-harut: It would be nice to have also what is the element which is defined in the sequence. In your case, it is x0 = 2
Do this in ALL tasks wherever needed.

PR #89, opened 2018-07-10
Hripsime: Tasks from 21 to 30

hsp-hripsime:


src/main/java/am/hsp/apps/seua/Task_25.java
```diff
@@ -0,0 +1,122 @@
...
+	 * Returns a fixed String expression answer1, if it's possible to build a
+	 * triangle, otherwise returns a fixed String expression answer2.
+	 * 
+	 * @param number1
+	 * @param number2
+	 * @param number3
+	 * @return
+	 */
+	public static String buildTriangle(int number1, int number2, int number3, String answer1, String answer2) {
+		int result = sumThreeNumbers(compareInequalityBetweenOneSide(number1, number2, number3),
```
hsp-harut: Hmm.. actually not a good way to go with. So in this case you are ALWAYS comparing all combinations for the sides and then you are getting the sum. But actually, there is no need to continue comparisons as soon as you will find any invalid side. So you need to have here some "AND" comparisons, and when one of them will be "false", that would be enough to stop further comparisons and return false for the method "isValidTriangle"

src/main/java/am/hsp/apps/seua/Task_25.java
```diff
@@ -0,0 +1,122 @@
...
+	/**
+	 * Returns 1, if the sum of the first two numbers is greater than the 3rd
+	 * number. Otherwise returns 0.
+	 * 
+	 * @param number1
+	 * @param number2
+	 * @param number3
+	 * @return
+	 */
+	public static int compareInequalityBetweenOneSide(int number1, int number2, int number3) {
```
hsp-harut: Based on the method name, it is not clear what your method is actually doing. But your comment is pretty good. So why to not have the method name similar to the comment? "isSumOfFirstTwoGreaterthanThird" - and of course you need to change also the return type of the method...

src/main/java/am/hsp/apps/seua/Task_26.java
```diff
@@ -0,0 +1,90 @@
...
+public class Task_26 {
+
+	public static void main(String[] args) {
+		int number1 = getUserInput("Please enter the first number: ");
+		int number2 = getUserInput("Please enter the second number: ");
+		int number3 = getUserInput("Please enter the third number: ");
+
+		System.out.println();
+
+		int outputIfContainsEvenNumber = getUserInput("Please enter the output if there is at least one even number: ");
```
hsp-harut: Same!

src/main/java/am/hsp/apps/seua/Task_26.java
```diff
@@ -0,0 +1,90 @@
...
+
+	/**
+	 * Returns answer1, if one of the three integers is even. Otherwise returns answer2.
+	 * 
+	 * @param number1
+	 * @param number2
+	 * @param number3
+	 * @return
+	 */
+	public static int isThereEvenNumberAmongThreeIntegers(int number1, int number2, int number3, int answer1,
```
hsp-harut: 1. You should have cohesive method, which will just return the case is it possible or not!
2. Method name could be simplified, why to not have just "isAnyNumberEven" or better "isEven"

src/main/java/am/hsp/apps/seua/Task_27.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Scanner;
+
+/**
+ * Print true, if the given three numbers are Arithmetic Progression. Otherwise
+ * print false.
+ * 
+ * @author hripsime
+ *
+ */
+public class Task_27 {
```
hsp-harut: very good!

src/main/java/am/hsp/apps/seua/Task_28.java
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	/**
+	 * Counts the quotient between two integers.
+	 * 
+	 * @param number1
+	 * @param number2
+	 * @return
+	 */
+	public static float division(int number1, int number2) {
```
hsp-harut: it is better to have double here...

src/main/java/am/hsp/apps/seua/Task_29.java
```diff
@@ -0,0 +1,145 @@
...
+	/**
+	 * Arranges the three numbers in ascending order.
+	 * 
+	 * @param number1
+	 * @param number2
+	 * @param number3
+	 * @return
+	 */
+	public static String arrangeInAscendingOrder(int number1, int number2, int number3) {
+		return (getMinOfThreeIntegers(number1, number2, number3) + ", "
```
hsp-harut: We discussed this case already

src/main/java/am/hsp/apps/seua/Task_30.java
```diff
@@ -0,0 +1,145 @@
...
+
+import java.util.Scanner;
+
+/**
+ * Print the three integers in descending order.
+ * 
+ * @author hripsime
+ *
+ */
+public class Task_30 {
```
hsp-harut: same

PR #103, opened 2018-07-26
Hripsime: Tasks 151-160

hsp-hripsime:


src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -317,4 +344,84 @@ protected String bubbleSortInDescendingOrder(int number1, int number2, int numbe
...
+		return sumOfFactors;
+	}
+
+	/**
+	 * Returns the product of the given number's factors.
+	 * 
+	 * @param number
+	 * @return
+	 */
+	protected long multiplyFactors(int number) {
```
hsp-harut: Same here. This is again some tasks specific functionality - and should be written in appropriate tasks.

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -317,4 +344,84 @@ protected String bubbleSortInDescendingOrder(int number1, int number2, int numbe
...
+	}
+
+	/**
+	 * Returns the sum of the two digit numbers, which are multiples of a given
+	 * number.
+	 * 
+	 * @param number
+	 * @return
+	 */
+	protected int sumTwoDigitMultiples(int number) {
```
hsp-harut: same is here.

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -317,4 +344,84 @@ protected String bubbleSortInDescendingOrder(int number1, int number2, int numbe
...
+		int sumOfMultiples = 0;
+
+		for (int multiple = 10; multiple <= 99; multiple++) {
```
hsp-harut: magic numbers.

src/main/java/am/hsp/apps/seua/tasks151_160/Task_153.java
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	/**
+	 * Returns the sum of those natural numbers, by which the given natural number is
+	 * divided leaving a fixed REMAINDER.
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getResult(int number) {
```
hsp-harut: Non cohesive method. You implemented 2 functionalities here:
1. Sum calculation
2. Checks whether each new number "i" is a divisor by balance (RMAINDER) for our number (don't you think this is a useful and generic functionality for any number related tasks ;) ).

src/main/java/am/hsp/apps/seua/tasks151_160/Task_153.java
```diff
@@ -0,0 +1,50 @@
...
+		int sumOfCertainDivisors = 0;
+
+		for (int i = 1; i <= number; i++) {
```
hsp-harut: What do you think, what is the possible maximum number which could be divisor by required balance?

src/main/java/am/hsp/apps/seua/tasks151_160/Task_154.java
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	/**
+	 * Returns the product of those natural numbers, by which the given natural number
+	 * is divided with the fixed REMAINDER.
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getResult(int number) {
```
hsp-harut: same issues as for the previous task.

src/main/java/am/hsp/apps/seua/tasks151_160/Task_156.java
```diff
@@ -0,0 +1,51 @@
...
+	 * given numbers.
+	 * 
+	 * @param number1
+	 * @param number2
+	 * @return
+	 */
+	private long getResult(int number1, int number2) {
+		long product = 1;
+
+		for (int i = 10; i <= 99; i++) {
```
hsp-harut: magic numbers

src/main/java/am/hsp/apps/seua/tasks151_160/Task_157.java
```diff
@@ -0,0 +1,48 @@
...
+	 * given number.
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getResult(int number) {
+		int sum = 0;
+
+		for (int i = 100; i <= 999; i++) {
+			if (!isMultiple(i, number)) {
```
hsp-harut: Please remind me to discuss this in our lesson.
Based on clean code principles it is a good practice to have a method for checking a negative case.
So if you will have a method isNotMultiplier - that would be completely clear for user/developer what is its purpose!

src/main/java/am/hsp/apps/seua/tasks151_160/Task_158.java
```diff
@@ -0,0 +1,49 @@
...
+	 * 
+	 * @param number1
+	 * @param number2
+	 * @return
+	 */
+	private long getResult(int number1, int number2) {
+		long product = 1;
+
+		for (int i = 100; i <= 999; i++) {
+			if (!isMultiple(i, number1) && !isMultiple(i, number2)) {
```
hsp-harut: please see my comment in the previous task.

src/main/java/am/hsp/apps/seua/tasks151_160/Task_159.java
```diff
@@ -0,0 +1,56 @@
...
+	 * divisors DIVISOR_1 and DIVISOR_2 with the fixed remainders REMINDER_1 and
+	 * REMINDER_2.
+	 * 
+	 * @param divisor1
+	 * @param divisor2
+	 * @param remainder1
+	 * @param remainder2
+	 * @return
+	 */
+	private double getResult(int divisor1, int divisor2, int remainder1, int remainder2) {
```
hsp-harut: Please see my comments in Task_153

src/main/java/am/hsp/apps/seua/tasks151_160/Task_159.java
```diff
@@ -0,0 +1,56 @@
...
+		double product = 1;
+
+		for (int i = 100; i <= 999; i++) {
```
hsp-harut: magic numbers

src/main/java/am/hsp/apps/seua/tasks151_160/Task_160.java
```diff
@@ -0,0 +1,46 @@
...
+		System.out.println("The result is: " + getResult());
+	}
+
+	/**
+	 * Returns the smallest three digit number, which has a natural square root when
+	 * multiplied by the given number.
+	 * 
+	 * @return
+	 */
+	private int getResult() {
```
hsp-harut: very very good

src/main/java/am/hsp/apps/seua/tasks151_160/Task_160.java
```diff
@@ -0,0 +1,46 @@
...
+		int i;
```
hsp-harut: hmm... Here is some kind of violation of SRP principle for "i" variable. 
So you are considering "i" to be used in the loop to store each next number after each iterations, and also you are considering "i" to be also required result for the overall method. So what if there will not be any number which has a natural root for the specified range of numbers (don't tell me that it is impossible - that there are definitely such numbers available in the range from 100 to 1000. As I said you have to design your methods to be independent of such dynamic data as a range of possible numbers)? So fir that case your method will return the last number in the specified range - which is not correct.

So you have 2 solutions here:
1. Crate a new variable "result" with a initial value equal to some constant lets say "private satic final int NOT_FOUND=-1"; and then return the result (as you did it for "i")
2. Just return "i" instead of "break", and final "return" of the method should return "NOT_FOUND" constant.

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -317,4 +344,84 @@ protected String bubbleSortInDescendingOrder(int number1, int number2, int numbe
...
+	 * 
+	 * @param number
+	 * @return
+	 */
+	protected int sumFactors(int number) {
+		if (number == 1) {
+			return 1;
+		}
+
+		int sumOfFactors = (number + 1);
```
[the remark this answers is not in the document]
hsp-hripsime: I wrote (number + 1), as all numbers are divided by 1 and themselves. So, there is no need to check if a number is divided by 1 without a remainder in the for() loop.

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -317,4 +360,56 @@ protected String bubbleSortInDescendingOrder(int number1, int number2, int numbe
...
+	
+	/**
+	 * Checks, if the given divisor is not the factor of the given number.
+	 * 
+	 * @param number
+	 * @param divisor
+	 * @return
+	 */
+	protected boolean isNotMultiple(int multiple, int number) {
+		return (multiple % number != 0);
```
hsp-harut: It would be cleaner to use already defined method with negative case: 
return !isMultiple

src/main/java/am/hsp/apps/seua/base/TaskBase.java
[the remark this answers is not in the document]
hsp-harut: hmm, I see. So in that case you need to write a comment on that line of code, so it will be clear what is the actual logic there.

src/main/java/am/hsp/apps/seua/base/TaskDigitBase.java
```diff
@@ -86,14 +74,14 @@ protected int getOneThousands(int number) {
 	protected int getMaxDigit(int number) {
 		int degree = 1;
 		int maxDigit = getOnes(number);
-		int result = number / getPower(10, degree);
+		int removeLastDigit = number / getPower(10, degree);
```
hsp-harut: 1. Actually bad variable name. Variable names should be nouns.
2. Why to not have a method which will remove last digit of the number and will return the result? Then you can use that method here? 
To remove the last digit, all you need is to divide the number to 10! So why to make the code more complicated by using here "pow" method?
3. Maybe it would be simpler if you will just modify actual number (remove last digit and assign result to number) rather than create a new variable?

src/main/java/am/hsp/apps/seua/base/TaskDigitBase.java
```diff
@@ -108,14 +96,14 @@ protected int getMaxDigit(int number) {
 	protected int getMinDigit(int number) {
 		int degree = 1;
 		int minDigit = getOnes(number);
-		int result = number / getPower(10, degree);
+		int removeLastDigit = number / getPower(10, degree);
```
hsp-harut: same

src/main/java/am/hsp/apps/seua/base/TaskDigitBase.java
```diff
@@ -131,17 +119,17 @@ protected int getMinDigit(int number) {
 	 */
 	protected boolean doesContainSpecifiedDigit(int number, int digit) {
 		int degree = 0;
-		int result = number;
+		int removeLastDigit = number;
```
hsp-harut: same

src/main/java/am/hsp/apps/seua/base/TaskDigitBase.java
```diff
@@ -155,11 +143,11 @@ protected boolean doesContainSpecifiedDigit(int number, int digit) {
 	 */
 	protected int countDigits(int number) {
 		int countDigits = 1;
-		int result = number / getPower(10, countDigits);
+		int removeLastDigit = number / getPower(10, countDigits);
```
hsp-harut: same

src/main/java/am/hsp/apps/seua/base/TaskDigitBase.java
```diff
@@ -174,12 +162,12 @@ protected int countDigits(int number) {
 	protected int sumDigits(int number) {
 		int degree = 1;
 		int sum = getOnes(number);
-		int result = number / getPower(10, degree);
+		int removeLastDigit = number / getPower(10, degree);
```
hsp-harut: same

src/main/java/am/hsp/apps/seua/base/TaskDigitBase.java
```diff
@@ -194,12 +182,12 @@ protected int sumDigits(int number) {
 	protected int multiplyDigits(int number) {
 		int degree = 1;
 		int product = getOnes(number);
-		int result = number / getPower(10, degree);
+		int removeLastDigit = number / getPower(10, degree);
```
hsp-harut: same

src/main/java/am/hsp/apps/seua/tasks151_160/Task_151.java
```diff
@@ -0,0 +1,52 @@
...
+	 * 
+	 * @param number
+	 * @return
+	 */
+	protected int sumFactors(int number) {
+		if (number == 1) {
+			return 1;
+		}
+
+		int sumOfFactors = (number + 1);
```
hsp-harut: Please write a comment on this line saying actual logic behind this summary result!

src/main/java/am/hsp/apps/seua/tasks151_160/Task_153.java
```diff
@@ -0,0 +1,51 @@
...
+	/**
+	 * Returns the sum of those natural numbers, by which the given natural number
+	 * is divided leaving a fixed REMAINDER.
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getResult(int number) {
+		int sumOfCertainDivisors = 0;
+		int maxDivisor = (number - REMAINDER);
```
hsp-harut: Yes, this makes clear what is behind the logic of subtraction! But even if you will not create a variable and just wrote this in the "for" loop - I will be ok with that also!

src/main/java/am/hsp/apps/seua/tasks151_160/Task_160.java
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	/**
+	 * Returns the smallest three digit number, which has a natural square root when
+	 * multiplied by the given number.
+	 * 
+	 * @return
+	 */
+	private int getResult() {
+		int result = NOT_FOUND;
```
hsp-harut: You can remove this variable, and just return "NOT_FOUND"

src/main/java/am/hsp/apps/seua/base/TaskBase.java
hsp-hripsime: Well, I thought it would not be a good practice to use a ! isMultiple() method)))

hsp-harut: no no :)) You need to use, otherwise, you will violate DRY principle!

Do not mix these 2 different concepts:
1. Use !isMultiple() anywhere you need this functionality (not a god practise)
2. Create a method for more convenient usage of "is not multiple" functionality, by implementing it as a negative case for already existed isMultiple() method :)

src/main/java/am/hsp/apps/seua/base/TaskDigitBase.java
```diff
@@ -172,37 +170,33 @@ protected int countDigits(int number) {
 	 * @return
 	 */
 	protected int sumDigits(int number) {
-		int degree = 1;
-		int sum = getOnes(number);
-		int result = number / getPower(10, degree);
+		int sumOfDigits = getOnes(number);
+		int result = removeLastDigit(number);
```
hsp-harut: same is here

src/main/java/am/hsp/apps/seua/base/TaskDigitBase.java
```diff
@@ -172,37 +170,33 @@ protected int countDigits(int number) {
...
 	 * 
 	 * @param number
 	 * @return
 	 */
 	protected int multiplyDigits(int number) {
-		int degree = 1;
-		int product = getOnes(number);
-		int result = number / getPower(10, degree);
+		int productOfDigits = getOnes(number);
+		int result = removeLastDigit(number);
```
hsp-harut: same

src/main/java/am/hsp/apps/seua/tasks151_160/Task_151.java
```diff
@@ -0,0 +1,54 @@
...
+	 * 
+	 * @param number
+	 * @return
+	 */
+	protected int sumFactors(int number) {
+		if (number == 1) {
+			return 1;
+		}
+
+		// All numbers' factors' sum is at least (number + 1), besides the
```
hsp-harut: 1. If you have several lines of comments, you need to use appropriate comment type for that /* */
2. I guess that would be great to have more descriptive and clear comment saying something like:
"Any number has at least 2 factors: actual number and 1. So to get rid of redundant iterations in the 'for' loop below, we are initializing summary result equal to 'number + 1'"

src/main/java/am/hsp/apps/seua/base/TaskDigitBase.java
```diff
@@ -154,15 +152,15 @@ protected boolean doesContainSpecifiedDigit(int number, int digit) {
 	 * @return
 	 */
 	protected int countDigits(int number) {
-		int countDigits = 1;
-		int result = number / getPower(10, countDigits);
+		int digitsQuantity = 1;
+		int result = removeLastDigit(number);
```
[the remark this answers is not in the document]
hsp-hripsime: I've not removed result parameter, as in that case number would become an output parameter.

Why to  initialize count to 0, if there is no number with digits of 0 quantity? At least they have 1 digit, that's why count = 1.  And as count = 1, I have had to initialize result outside of the loop to not enter the loop in case of one digit numbers. 

But if this complicates the method and there is no any other reason I can simplify the code by putting "result" inside the loop.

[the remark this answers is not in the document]
hsp-harut: @hsp-hripsime jan yes as I said I have no concerns with the "result" variable, and as I mentioned I have no concerns only if the reason for that is to get rid of "output param" issue for the number. That is pretty good that you designed your method that way!

Related you second concern regarding count initialization. Yes, any number has at least one digit, but actually even if you are initializing count equal to 1, you are not reducing the count of "remove last digit of the number" functionality, you just 'moved' one iteration of the while loop outside from it :) So this just makes the code to be complicated, as what is the reason for that one hardcoded iteration, if that is completely covered by 'while loop'? Why to duplicate the code, if it could be used only in the loop and make the required functionality?

Related the concern that currently you had to check the case result = 0 separately and return 1 for that case.
So this is a very good case when you can think about most preferable loop to use :) You can use "do/while" loop and that will solve everything - and there will be no need to check that case at the very beginning of the method.

PR #105, opened 2018-07-30
Hripsime: Tasks 161 -170

hsp-hripsime:


src/main/java/am/hsp/apps/seua/tasks161_170/Task_169.java
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * Returns 5, if the sum of the given two numbers is prime, otherwise
+	 * returns 6.
+	 * 
+	 * @param number1
+	 * @param number2
+	 * @return
+	 */
+	private int getResult(int number1, int number2) {
```
hsp-harut: We agreed that we can use "getResult" as a method name when the task requirement is too complex, and it is too hard to specify a method name for it, as it might contain a pretty much number of characters.
Actually, whenever we can create a method with a descriptive and meaningful name (same time not too long) - we should do it!

PR #110, opened 2018-08-02
Arman: Tasks_161-170

hsp-arman:
Tasks - 161-170


src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -222,4 +222,90 @@ protected int getPow(int number, int degree) {
...
+		return !isMultiple(number, divisor);
+	}
+	
+	/**
+	 * return "true" , if number is prime 
+	 * @param number
+	 * @return
+	 */
+	protected boolean isPrime(int number) {
+		for (int i = 2; i < number; i++) {
```
hsp-harut: what is the maximum possible integer that our number could be divided without any remainder?

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -222,4 +222,90 @@ protected int getPow(int number, int degree) {
...
+
+	/**
+	 * return "true" , if the number has a natural number root
+	 * 
+	 * @param number
+	 * @return
+	 */
+	protected boolean isRoot(int number) {
+		for (int i = 1; i <= number / 2; i++) {
+			if (getPow(i, 2) == number) {
```
hsp-harut: we can improve your code even more. So why to continue with "for" loop iterations, when getPow(i, 2) is greater than number? So for that case the method must return false, and there is no need to continue with "for" iterations!

So having all mentioned above, I guess "for" loop is not the best one to use in this method.
You need to raise 'i' to power of 2 while the result is less than number yes? So when your loop will over, you will definitely know that the result is:
1. equal to number
2. greater than the number

So you just need to return "result == number" :)

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_164.java
```diff
@@ -0,0 +1,70 @@
+package main.java.am.hsp.apps.seua.tasks_161_170;
+
+import main.java.am.hsp.apps.seua.base.TaskDigit;
+
+/**
+ * Print smallest tree-digit number, which is bigger than the square of given
```
hsp-harut: Wrong comment. Based on your comment it seems that we need to find the minimum 3 digit number which is greater than the square root of given number: "aysinqn grel es gtnel ayn amenapoqr eranish tiv@ vor aveli mec e qan trvac tvi armat@"

But actually your task is "find the minimum 3 digit number which has a square root greater than given number"

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_164.java
```diff
@@ -0,0 +1,70 @@
...
+ * 
+ * @author arman
+ *
+ */
+public class Task_164 extends TaskDigit {
+
+	private static final int START_NUMBER = 100;
+	private static final int FINISH_NUMBER = 999;
+	private static final int NO_SUCH_NUMBER = 0;
+	private static final int BASE = 2;
```
hsp-harut: POWER

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_164.java
```diff
@@ -0,0 +1,70 @@
...
+
+	/**
+	 * return smallest tree-digit number, which is bigger than the square of given
+	 * number
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getSmallestNumberWhichIsBiggerThanSquareToGivenNumber(int number) {
+		return getPow(number, BASE) < START_NUMBER ? getFirstSquareNumber(START_NUMBER) : getPow(number + 1, BASE);
```
hsp-harut: Wrong!!!

1. You checked if getPow(number, 2) < 100 then return getFirstSquareNumber??? Why?? For this case it is clear that 100 (START_NUMBER) is the minimum 3 digit number which has a square root greater than given number!!!

2. You wrote if getPow(number, 2) >= 100, then you are returning the pow of the next number (number + 1) ??? So you think that if my number = 11 and 121 > 100, then your desired number is 144 ? No of course, your desired number for this case is 122!!!

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_164.java
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	/**
+	 * return smallest tree-digit number, which is bigger than the square of given
+	 * number
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getSmallestNumberWhichIsBiggerThanSquareToGivenNumber(int number) {
```
hsp-harut: Very long and wrong (it is wrong for the same reason I wrote for the class comment) method name.
getMinNumberWithSquareRootGreaterThanGiven

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_164.java
```diff
@@ -0,0 +1,70 @@
...
+		return startNumber > FINISH_NUMBER ? NO_SUCH_NUMBER : startNumber;
+	}
+
+	/**
+	 * return first 3-digit square number
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getFirstSquareNumber(int number) {
```
hsp-harut: So you are finding the first number which has a square root as Integer value (see implementation of isRoot method). But I guess this is not what is required in your task. 
The square root of the number can be also a Double - and that is ok for this task. 
So I guess this method is not needed at all. And in this task you can use **Math.sqrt** method for getting the square root of the number.

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_166.java
```diff
@@ -0,0 +1,45 @@
+package main.java.am.hsp.apps.seua.tasks_161_170;
+
+import main.java.am.hsp.apps.seua.base.TaskDigit;
+
+/**
+ * Print "ANSWER_TRUE" if given number is multiple to DIVISOR , else
```
hsp-harut: same

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_166.java
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * return "ANSWER_TRUE" if given number is multiple to DIVISOR , else
+	 * "ANSWER_FALSE"
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getBooleanToIntIsMultiple(int number) {
+		return !isDegree(number, BASE) ? ANSWER_TRUE : ANSWER_FALSE;
```
hsp-harut: No. Based on your task description you need to print "0" when !isDegree(number, BASE)

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_166.java
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	/**
+	 * return "ANSWER_TRUE" if given number is multiple to DIVISOR , else
+	 * "ANSWER_FALSE"
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getBooleanToIntIsMultiple(int number) {
```
hsp-harut: Bad and complicated method name. So this method just returns the actual final result what should be printed.
So just have here "getResult"

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_167.java
```diff
@@ -0,0 +1,48 @@
+package main.java.am.hsp.apps.seua.tasks_161_170;
+
+import main.java.am.hsp.apps.seua.base.TaskDigit;
+
+/**
+ * Print "true" if is even negative value , else "false"
```
hsp-harut: Wrong description. You didn't even mention "sin" here...

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -273,4 +276,53 @@ protected boolean isMultiple(int number, int divisor) {
...
+	 * @return
+	 */
+	protected boolean isPrime(int number) {
+		if (number == 2) {
+			return true;
+		}
+		if (!isOdd(number)) {
+			return false;
+		}
+		for (int i = 3; getPow(i, 2) < number; i += 2) {
```
hsp-harut: vay Arman jan :)))) Yndamney@ petq er minchev number/2 gnair loopid mej u skseir i = 2 ic

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -273,4 +276,53 @@ protected boolean isMultiple(int number, int divisor) {
...
+	/**
+	 * return "true" , if number is prime 
+	 * @param number
+	 * @return
+	 */
+	protected boolean isPrime(int number) {
+		if (number == 2) {
+			return true;
+		}
+		if (!isOdd(number)) {
```
hsp-harut: not needed, as this violated KISS and makes your code complicated. This checking will be done in the first "for" iteration, so writing this here will not make your program to have better performance - this is not the case...

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -243,14 +243,17 @@ protected boolean isMultipleWithRemainder(int number, int divisor, int remainder
 	 * @return
 	 */
 	protected boolean isRoot(int number) {
-		for (int i = 1; i <= number / 2; i++) {
-			if (getPow(i, 2) == number) {
+		int j = 1;
+		for (int i = 1; i <= number; i = getPow(j, 2)) {
```
hsp-harut: No Arman jan, you didn't followw my comment. I mentioned that the "for" loop is not the right loop for this functioanlity.
And I tried to describe you an algorithm. So based on my description you need to use a "do/while" loop. 
Arman jan if you have some difficulties on getting my idea in the comments - always contact me for the description!

Now let me here copy past my previous comment:
**_You need to raise 'i' to power of 2 while the result is less than number! So when your loop will over, you will definitely know that the result is:

a)   equal to number
b)   greater than the number

So you just need to return "result == number"_**

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_164.java
```diff
@@ -0,0 +1,54 @@
...
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getMinNumberWithSquareRootGreaterThanGiven(int number) {
+		double startNumber = (double) START_NUMBER;
+		while (startNumber < FINISH_NUMBER) {
+			if (Math.sqrt(startNumber) > number) {
+				return (int) startNumber;
+			} else {
```
hsp-harut: do you need "else" block? What if you will remove it?

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_164.java
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	/**
+	 * return the minimum 3 digit number which has a square root greater than given
+	 * number
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getMinNumberWithSquareRootGreaterThanGiven(int number) {
```
hsp-harut: As we discussed in the slack you found much better solution without using any loop.
Please write a new method (DON'T REMOVE THIS ONE!!!) by that solution!

src/main/java/am/hsp/apps/seua/tasks_161_170/Task_166.java
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * return "ANSWER_FALSE" if given number is not degree to base , else
+	 * "ANSWER_TRUE"
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getResult(int number) {
+		return !isDegree(number, BASE) ? ANSWER_TRUE : ANSWER_FALSE;
```
hsp-harut: You didn't fix my comment. Your program will show incorrect result. It should print 0 when the number is NOT degree, but in your program it will print "1"

src/main/java/am/hsp/apps/seua/base/TaskBase.java
hsp-arman: ok , hesa kuxxem , bayc senc loop@  shat karch a frum ))

src/main/java/am/hsp/apps/seua/base/TaskBase.java
hsp-arman: (Y)

src/main/java/am/hsp/apps/seua/base/TaskBase.java
hsp-harut: karch fral@ chapanish chi! - ba ed pow-n hashveluc qani gorcoxutyuna anum da ches hashvum ? :))

hsp-harut: chgitem karoxa chisht es - bayc aveli lava KISS-@ pahenq - parz u hasarak - es en gorcoxutyun@ chi vor inch vor dandaxutyan beri vor porcenq optimalacnenq !!

PR #113, opened 2018-08-17
Hripsime: Tasks 171 - 180

hsp-hripsime:


src/main/java/am/hsp/apps/seua/tasks171_180/Task_178.java
```diff
@@ -0,0 +1,132 @@
...
+			System.out.print(ONE_SPACE);
+			System.out.print(SECOND_MEMBER);
+		}
+
+		if (index > 2) {
+			System.out.print(ONE_SPACE);
+			System.out.print(THIRD_MEMBER);
+		}
+
+		for (int i = 1; i <= index - 3; i++) {
```
hsp-harut: hmm... I guess I understood why you are iterating until index - 3. So you printed first 3 members before the loop, and thats why you are doing 3 times less iterations in the for loop. So this is a good sample where you can see that the code is quite a bit complicated, and I get the logic only in this task!

Actually, even if you are not using "i" in the calculations, BUT based on method context "i" corresponds to the INDEX of the element in the Sequence! So that would be correct if you will start your "for" iterations starting from the Index which you didn't print (actually didn't print element by that index) yet! So in your case you need to start the loop from i = 4 and will continue until the final index!

What if you will need to print also element index before its value ? :) So your "i" should be the index of the element - which you can use or not in the further logic!

PR #114, opened 2018-08-22
Arman: Tasks_181-190

hsp-arman:
Arman: Tasks_181-190


src/main/java/am/hsp/apps/seua/tasks_181_190/Task_185.java
```diff
@@ -0,0 +1,67 @@
...
+		getMoneyByPercent(number);
+	}
+
+	/**
+	 * print money in every month and expiration month number number
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private void getMoneyByPercent(int percent) {
```
hsp-harut: If your method purpose is print any data, then the method name should not start with "get" (get - means method must return any value!)!

src/main/java/am/hsp/apps/seua/tasks_181_190/Task_188.java
```diff
@@ -0,0 +1,58 @@
...
+
+	/**
+	 * return "true" , if given number is number Fibonacci ,else "false"
+	 * 
+	 * @param number
+	 */
+	private boolean isNumberFibonacci(int number) {
+		int i = 1;
+		
+		while (getFibonacci(i) < number) {
```
hsp-harut: You are using a while loop and that's why in the "return" part you are calculating once again fibonacci for the same 'i' which you already calculated in the "while". So this is a good example that you use a wrong loop type.
If you will use "do/while" loop, then you can assign getFibonacci(i) to any local variable (declared outside of "do" block) which you can also use in the "return" part comparison...

src/main/java/am/hsp/apps/seua/tasks_181_190/Task_190.java
```diff
@@ -0,0 +1,77 @@
...
+		return (getElementByIndex(index - 2) + 2 * getElementByIndex(index - 1)); // 3;
+	}
+
+	/**
+	 * Print minimum value of k , if Math.abs(x[k] - x[k-1] is smaller than the
+     * given number , and x[k] = (x[k-2] + 2x[k-1]) / 3 ; x[1] = 1 ; x[2] = 2
+	 * 
+	 * @param number
+	 */
+	private int printResult(double number) {
```
hsp-harut: based on some rules we learned in previous lesson related clean code conventions, it is better that you will declare this method AFTER "solve" method - as this is the method which is used in the "solve" method first... 
You can also check you previous tasks in this PR and reorder methods declarations based on their order of usage...

hsp-arman:
Dear Harut, Tasks 186 and 187 are not solved․ Task 186 is very similar in Task 185 ,  and Task 187 is very easy.

src/main/java/am/hsp/apps/seua/tasks_181_190/Task_190.java
```diff
@@ -0,0 +1,73 @@
...
+		int i = 2;
+
+		while (getComparisonCaseForElemByIndex(i) < number) {
```
hsp-harut: You need to find the smallest index for which this expression is true. But based on your current implementation you are trying to ding the smallest index for which this expression is false - which is incorrect.

src/main/java/am/hsp/apps/seua/tasks_181_190/Task_188.java
```diff
@@ -0,0 +1,52 @@
...
+		System.out.println(isFibonacci(number));
+	}
+
+	/**
+	 * return "true" , if given number is number Fibonacci ,else "false"
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private boolean isFibonacci(int number) {
```
hsp-harut: Not cohesive method: a) You are calculating fibonacci number b) You are checking whether number is a fibonacci number or not.

I don't understand why you overwrited your previous implementation of this method - which was VERY clean and correct...

src/main/java/am/hsp/apps/seua/tasks_181_190/Task_188.java
```diff
@@ -0,0 +1,52 @@
...
+		int prePrevious = PRE_PREVIOUS;
+		int previous = PREVIOUS;
+		int current = 1;
+
+		for (int i = 2; i <= number; i += prePrevious) {
```
hsp-harut: I don't know why you decided to use for loop for this task. Based on your task it is not clear how many iterations you need to do untill you will get the desired result. So that's why "for" loop is not the best approach.
As I said your previous solution was pretty good, and you asked you to make a small enhancements. But you decided to remove everything and rewrite the method - and I don't know why ... 
Please undo your changes to your previous solution and just fix a small issue I wrote in my previous review

src/main/java/am/hsp/apps/seua/tasks_181_190/Task_189.java
```diff
@@ -0,0 +1,66 @@
...
+	}
+	
+
+	/**
+	 * return "true" , if given number is number Fibonacci ,else "false"
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private boolean isFibonacci(int number) {
```
hsp-harut: Same

PR #115, opened 2018-08-22
Lusine: Tasks 181_ 190

hsp-lusine:


src/main/java/am/hsp/apps/seua/task181_190/Task_187.java
```diff
@@ -0,0 +1,25 @@
+package main.java.am.hsp.apps.seua.task181_190;
+/**
+ * Print "true", if the given number Prime, otherwise Print "false"
+ */
+import main.java.am.hsp.apps.seua.base.TaskBase;
+
+public class Task_187 extends TaskBase {
```
hsp-harut: As I said no need to implement the tasks which are the same as your previous tasks.

src/main/java/am/hsp/apps/seua/task181_190/Task_190.java
```diff
@@ -0,0 +1,77 @@
...
+	/**
+	 * Decide the smallest number N, in which case |x[n] - x[n-1]| < Epsilion
+	 * @param epsilion
+	 * @return
+	 */
+	private int getResult(int epsilion) {
+		int index = 1;
+		double result = FIRST_NUMBER;
+
+		while (result < epsilion) {
```
hsp-harut: You need to find the smallest index for which this expression is true. But based on your current implementation you are trying to ding the smallest index for which this expression is false - which is incorrect.

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -338,17 +359,30 @@ protected long getFactorial(int number) {
...
 		return product;
 	}
 
+	/**
+	 * Calculate the current Fibonacci Number with Recursion F[n-1] + F[n-2]
+	 * 
+	 * @param index
+	 * @return
+	 */
+	protected int getFibonacciNumbers(int index) {
```
hsp-harut: You are getting many numbers or just one? I guess only one. In that case update your method name - it does not return numberS, it returns a number

src/main/java/am/hsp/apps/seua/task181_190/Task_182.java
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	/**
+	 * Calculate the maximum number K, the square of which is less or equal than the
+	 * specified number N
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private int getBiggestBase(int number) {
```
hsp-harut: You wrote in the comment "maximum" and you named your method "biggest"? hmm... too strange. Max and Min much more preferable for this context

src/main/java/am/hsp/apps/seua/task181_190/Task_186.java
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * The skier has passed WAY.Each next day it is added P Percent. Calculate when
+	 * the WAY will be greater than FINAL_WAY
+	 * 
+	 * @param speed
+	 * @param percent
+	 */
+	private void printFinishWayAndDay(double way, int percent) {
+		double currentWayofDay = WAY;
+		int day = 0;
```
hsp-harut: First day skier passed 20 km distance. So if you initialized allCurrentWay = 20, then why your day number = 0?

src/main/java/am/hsp/apps/seua/task181_190/Task_187.java
hsp-harut: Where is the comment on the class name???

PR #131, opened 2018-09-03
Hripsime: Tasks 251-260

hsp-hripsime:
The left-hand side tasks have the same logic as the right-hand side tasks, that's why they're missing:
254 --> 253
256 --> 255
259 --> 257
260 --> 258


src/main/java/am/hsp/apps/seua/tasks251_260/Task_255.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	/**
+	 * Returns the sum of the max element of the given array and its' index. If
+	 * there are several max elements, it gets the first one.
+	 * 
+	 * @param array
+	 * @return
+	 */
+	private int sumMaxElementAndItsIndex(int[] array) {
```
hsp-harut: get

PR #140, opened 2018-09-06
Hripsime: Tasks 281-320

hsp-hripsime:


src/main/java/am/hsp/apps/seua/tasks281_320/Task_320.java
```diff
@@ -0,0 +1,83 @@
...
+	 * 
+	 * @param array
+	 */
+	private int[] getRequiredArray(int[] array) {
+		int firstMax = array[0];
+		int lastMin = array[0];
+		int firstMaxIndex = 0;
+		int lastMinIndex = 0;
+
+		for (int i = 0; i < array.length; i++) {
```
hsp-harut: not a pretty solution. Your method is too big and non cohesive.
You can have a method which will return an array with 2 elements containing required max and min indexes.
Then you can use that array in further checkings...

src/main/java/am/hsp/apps/seua/tasks281_320/Task_320.java
```diff
@@ -0,0 +1,96 @@
...
+	 * 
+	 * @param array
+	 */
+	private int[] getRequiredArray(int[] array) {
+		int firstMaxElement = array[0];
+		int lastMinElement = array[0];
+		int firstMaxIndex = 0;
+		int lastMinIndex = 0;
+
+		for (int i = 0; i < array.length; i++) {
```
hsp-harut: hmm... why do not extract this into a separate method which will return an array with the two elements containing required indexes? From clean code prespective, this method is not as clean as it could be.

src/main/java/am/hsp/apps/seua/tasks281_320/Task_318.java
```diff
@@ -0,0 +1,89 @@
...
+	 * the negative values, then 0s, then the positive values.
+	 * 
+	 * @param array
+	 * @return
+	 */
+	private int[] getRequiredArray(int[] array) {
+		int lastIndexForNegativeElements = -1;
+		int lastIndexForZeroElements = array.length - 1;
+
+		for (int i = 0; i < array.length; i++) {
```
hsp-harut: Please update according our discussion

PR #142, opened 2018-09-11
Hripsime: Tasks 321-350

hsp-hripsime:
In the right hand side are tasks chosen from the range.
(321 - 330) --> 326, 328, 329
(331 - 340) --> 334, 335, 337, 340
(341 - 350) --> 341, 344, 350


src/main/java/am/hsp/apps/seua/tasks321_350/Task_334.java
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	/**
+	 * Returns the index of the first maximum element of the given array, the sum of
+	 * positive elements and the product of the negative elements.
+	 * 
+	 * @param array
+	 * @return
+	 */
+	private int[] getIndexOfFirstMaxSumOfPositiveAndProductOfNegativeElements(int[] array) {
```
hsp-harut: Even your method name says that it is not cohesive!
Task says: 
that you need to calculate sum of positives Only if something = true
that you need to calculate product of negatives Only if something = false

So your algorithm skeleton must have this strucure:
getRequredResult() {
if(maximum index is less than number)
return calculateSum
} 

return calculateProduct;
}

hsp-harut: I know where you get this misunderstanding! In our previous lesson I suggested you to use a method which will return first max and last min indexes (if I am not mistaken).
But never mix! In that task your overal algorithm skeleton was something like:
if(any condition) {
return calculateSumOfMinAndMaxIndexes;
}

return someOtherCalculation;

So here you can see that "calculateSumOfMinAndMaxIndexes" this is separate and independent functionality in scope of the task - and it is ok to calculate both min and max indexes in scope of that method - which is again - cohesive functionality from the task prespective.

I hope I could explain what I mean, but we will have discussions on this during our lesson.

src/main/java/am/hsp/apps/seua/tasks321_350/Task_335.java
```diff
@@ -0,0 +1,80 @@
...
+		printRequiredElementAndIndex(array);
+	}
+
+	/**
+	 * Prints those element and its index of the given array, if the sum of its
+	 * previous and next elements are the greatest.
+	 * 
+	 * @param array
+	 */
+	private void printRequiredElementAndIndex(int[] array) {
```
hsp-harut: non cohesive method. 1. It does printing functionality 2. also implements some task specific logic.

src/main/java/am/hsp/apps/seua/tasks321_350/Task_335.java
```diff
@@ -0,0 +1,80 @@
...
+		if (isArrayLengthLessThanRequired(array)) {
+			System.out.println("\nThe array is too short to get the required result.");
+			return;
+		}
+
+		int requiredElement = array[1];
+		int requiredIndex = 1;
+
+		for (int i = 2; i < array.length - 1; i++) {
```
hsp-harut: You need to have a separate method which will return an index of desired element...

src/main/java/am/hsp/apps/seua/tasks321_350/Task_335.java
```diff
@@ -0,0 +1,80 @@
...
+			if (getSumOfPreviousAndNextElements(array, i) > getSumOfPreviousAndNextElements(array, i - 1)) {
+				requiredElement = array[i];
+				requiredIndex = i;
+			}
+		}
+
+		System.out.println("\nThe required element is: " + requiredElement);
```
hsp-harut: Why you decided to extract the functionality of communication with user into this method? Then what is the main purpose of "solve" method?

src/main/java/am/hsp/apps/seua/tasks321_350/Task_335.java
```diff
@@ -0,0 +1,80 @@
...
+	 * @see main.java.am.hsp.apps.seua.base.TaskBase#solve()
+	 */
+	public void solve() {
+		int[] array = getUserIntArrayInput("Please enter the array length: ", "Please enter the start of the range: ",
+				"Please enter the end of the range: ");
+
+		System.out.print("\nThe array is: ");
+		printIntArray(array);
+
+		printRequiredElementAndIndex(array);
```
hsp-harut: You need to get required index here, and then print element with index in solve method.

src/main/java/am/hsp/apps/seua/tasks321_350/Task_335.java
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	/**
+	 * Prints those element and its index of the given array, if the sum of its
+	 * previous and next elements are the greatest.
+	 * 
+	 * @param array
+	 */
+	private void printRequiredElementAndIndex(int[] array) {
+		if (isArrayLengthLessThanRequired(array)) {
```
hsp-harut: User communication should be done in "solve" method

src/main/java/am/hsp/apps/seua/tasks321_350/Task_337.java
```diff
@@ -0,0 +1,79 @@
...
+	/**
+	 * Returns the sum of the given array's elements with odd indexes, if the
+	 * product of the elements with even indexes is positive, otherwise returns
+	 * the given value.
+	 * 
+	 * @param array
+	 * @return
+	 */
+	private int getRequiredCalculation(int[] array) {
+		return (getProductOfElementsWithEvenIndexes(array) > 0) ? getSumOFElementsWithOddIndexes(array) : GIVEN_VALUE;
```
hsp-harut: It would be more clean and if you will have a method something like  **isProductOfEvenIndexElemsPositive**

src/main/java/am/hsp/apps/seua/tasks321_350/Task_340.java
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * Returns the minimum value of those elements, which indexes divided by the
+	 * given divisor, will remain the given remainder.
+	 * 
+	 * @param array
+	 * @return
+	 */
+	private void printRequiredMinValue(int[] array, int divisor, int remainder) {
+		if (isArrayLengthLessThanRequired(array, remainder)) {
```
hsp-harut: non cohesive method :( please see my comments on one of previous tasks

src/main/java/am/hsp/apps/seua/tasks321_350/Task_340.java
```diff
@@ -0,0 +1,69 @@
...
+			System.out.println("\nThe array length isn't enough to find the required value.");
+			return;
+		}
+
+		int minValue = array[remainder];
+
+		for (int i = (remainder + divisor); i < array.length; i += divisor) {
```
hsp-harut: hmm... but why? You implemented a lot of tasks and you always have your methods as cohesive!
But now in this method you are a) validating and printing validation message to user b) calculating required min value c) printing it to user.

src/main/java/am/hsp/apps/seua/tasks321_350/Task_341.java
```diff
@@ -0,0 +1,86 @@
...
+import main.java.am.hsp.apps.seua.base.TaskArrayBase;
+
+/**
+ * Print the given array's elements quantity, which have no equal values in the
+ * array.
+ * 
+ * @author user
+ *
+ */
+public class Task_341 extends TaskArrayBase {
```
hsp-harut: Please see my comments in similar task.

src/main/java/am/hsp/apps/seua/tasks321_350/Task_350.java
```diff
@@ -0,0 +1,83 @@
...
+
+	/**
+	 * Returns an array, which first element is the sum of positive elements and the
+	 * second one is the sum of the last 5 positive elements raised to the given
+	 * power.
+	 * 
+	 * @param array
+	 * @return
+	 */
+	private int[] getRequiredResults(int[] array) {
```
hsp-harut: Why do not return required avarage value in scope of this method? Returning type of array of integers makes this method complicated. I see no reason why to not have here required calculation and then return it?

src/main/java/am/hsp/apps/seua/tasks321_350/Task_350.java
```diff
@@ -0,0 +1,83 @@
...
+		int[] requiredResults = new int[] { 0, 0 };
+		int index = array.length;
+
+		while (isConditionSatisfiedForSearchingNextElement(requiredResults[0], --index)) {
```
hsp-harut: hmm... I don't know... For me, this is too complicated. You know, this just makes the reader to be scared! It is too abstract "isConditionSatisfiedForSearchingNextElement" and I don't know what to expect from this method! I am imagining that there should be some complicated and hard logic :)))

Anyway, for me, it is much easier and predictable if the while will contain only the logic which will say (I don't know how to explain) how many times to iterate - maybe the possible maximum iteration number! Then in the while loop we can check our task specific checking related positive elements count! And if the count reached - then we can just "return"!
You know that way is much simpler to get actual logic behind the "while" and overall functionality.

src/main/java/am/hsp/apps/seua/tasks321_350/Task_334.java
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/**
+	 * Checks whether the first maximum element's index is less than or equal to the
+	 * given value.
+	 * 
+	 * @param array
+	 * @return
+	 */
+	private boolean isConditionSatisfied(int[] array, int number) {
```
hsp-harut: 1. I guess we could have here a descriptive name, so why not to have?
isIndexOfFirstMaxNotGreaterThanNumber

src/main/java/am/hsp/apps/seua/tasks321_350/Task_334.java
```diff
@@ -0,0 +1,151 @@
...
+	 * Returns the positive elements sum of the given array, if the index of the
+	 * first maximum element is not more than the given number, otherwise returns
+	 * the product of the negative elements.
+	 * 
+	 * @param array
+	 * @param number
+	 * @return
+	 */
+	private int getRequiredResult(int[] array, int number) {
+		if (isConditionSatisfied(array, number)) {
```
hsp-harut: This is small issue, but this is more than important in real life - to follow actual requirement. The task says that you need to compare first max index with "n / 2" (not with 'n')

src/main/java/am/hsp/apps/seua/tasks321_350/Task_334.java
```diff
@@ -0,0 +1,151 @@
...
+
+	/**
+	 * Returns the product of the negative elements of the given array. If there is
+	 * no negative elements, returns NOT_FOUND.
+	 * 
+	 * @param array
+	 * @return
+	 */
+	private int getProductOfNegativeElements(int[] array) {
+		if (doesContainNegativeElement(array)) {
```
hsp-harut: but why? there is no such requirement. This calls YAGNI! And because of this checking you created a new method for calculating required product, and as the descriptive name you already used here, you named that 'main' method as "getRequiredProduct" - which is too abstract.

I understand why you checked this - to not return "1"  when there will be no negative elements in the array.
But anyway, if you decided to consider that case, it is much better to initialize sum = 0, and in each iteration check and initialize sum = 1, rather than have a separate method which will have "n" complexity (worst case will make "n" number of iterations).

Anyway, it is very good that you are considering such cases! Keep this way!

src/main/java/am/hsp/apps/seua/tasks321_350/Task_335.java
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	/**
+	 * Returns the element from the given array by the given index.
+	 * 
+	 * @param array
+	 * @param index
+	 * @return
+	 */
+	private int getElementByIndex(int[] array, int index) {
```
hsp-harut: I guess no need for this method, this does nothing than array[i] does. Doesn't it?

src/main/java/am/hsp/apps/seua/tasks321_350/Task_341.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	/**
+	 * Checks whether the given element is unique.
+	 * 
+	 * @param array
+	 * @param element
+	 * @return
+	 */
+	private boolean isElementUnique(int[] array, int element) {
```
hsp-harut: Maybe it is good idea to place this in corresponding base class? This is useful method, and that would be nice to have this in base class (also we had similar logic implemented in other tasks also)!

PR #149, opened 2018-09-21
Lusine: Searching_Algorithms

hsp-lusine:


src/main/java/am/hsp/apps/algorithms/search/LinearSearch.java
```diff
@@ -1,36 +1,80 @@
...
-	private int getIndexElmEqualNumber(int[] array, int number) {
-		for (int i = 0; i < array.length; i++) {
-			if (array[i] == number) {
-				return i;
-			}
-		}
-
-		return NOSOLUTION;
-	}
+public class LinearSearch extends SearchAlgorithms {
```
hsp-harut: Same as fr previous tasks

src/main/java/am/hsp/apps/algorithms/search/SearchAlgorithms.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Check number is in range
+     * 
+     * @param number
+     * @param begin
+     * @param end
+     * @return
+     */
+    protected boolean isNumberOfRange(int number, int begin, int end) {
```
hsp-harut: I guess this method is more generic and should be moved to appropriate class

src/main/java/am/hsp/apps/algorithms/search/InterpolationSearch.java
```diff
@@ -0,0 +1,117 @@
...
+     * 
+     * @see main.java.am.hsp.apps.algorithms.search.SearchAlgorithms#search(int[],
+     * int)
+     */
+    @Override
+    public int search(int[] array, int number) {
+        int beginIndex = 0;
+        int endIndex = array.length - 1;
+
+        if (isNumberOutOfRange(number, array[beginIndex], array[endIndex])) {
```
hsp-harut: If you have such validation here, why you didn't use is in other searches (for example in binary search)?

src/main/java/am/hsp/apps/algorithms/search/InterpolationSearch.java
```diff
@@ -0,0 +1,117 @@
...
+            return NOT_SOLUTION;
+        }
+
+        while (beginIndex < endIndex && array[endIndex] != array[beginIndex]) {
```
hsp-harut: As we tested together, the infinity loop problem still exists (for the case: number = 170)
So to fix the issue, we need to go by the way Hripsime suggested, hust do the same validation isNumberOutOfRange in each while iteration.
But as you also have another validation checking in the while loop: array[endIndex] != array[beginIndex], it would be better to create a new method lets say: isValidToContinue, and place there both validations you need to have in the "while" iterations

PR #150, opened 2018-09-22
Arman: Searching Algorithms

hsp-arman:
Searching Algorithms :
Linear Search;
Binary Search;
Jump Search;
Interpolation Search.


src/main/java/am/hsp/apps/algorithms/search/BinarySearch.java
```diff
@@ -1,26 +1,97 @@
...
+        while (startIndex <= endIndex) {
+            int midIndex = getMidIndex(startIndex, endIndex);
+
+            if (number < array[midIndex]) {
+                endIndex = midIndex - 1;
+            } else if (number > array[midIndex]) {
+                startIndex = midIndex + 1;
+            } else {
+
+                return midIndex;
```
hsp-harut: It would be cleaner and more predictable if you will check this case in the very beginning of the while loop - when you get a middle element index you need to checks whether you found the number - and if yes, then return the index!
And only after that - you need to proceed with the functionality for re-initializing  start/end indexes

src/main/java/am/hsp/apps/algorithms/search/BinarySearch.java
```diff
@@ -1,26 +1,97 @@
...
+    /**
+     * Return result by binary search with a recursive method
+     * 
+     * @param array
+     * @param startIndex
+     * @param endIndex
+     * @param number
+     * @return
+     */
+    public int getResultByBinarySearchRecursive(int[] array, int startIndex, int endIndex, int number) {
```
hsp-harut: Why just not searchByRecursion or searchRecursively ?

src/main/java/am/hsp/apps/algorithms/search/BinarySearch.java
```diff
@@ -1,26 +1,97 @@
...
+        int midIndex = getMidIndex(startIndex, endIndex);
 
-        int[] temporaryArray = { 10, 20, 80, 30, 60, 50, 110, 100, 130, 170 };
+        if (endIndex < startIndex) {
```
hsp-harut: hmm... I don't understand you are first getting midIndex, and only then you are checking whether start/end indexes are valid or not?

src/main/java/am/hsp/apps/algorithms/search/BinarySearch.java
```diff
@@ -1,26 +1,97 @@
...
+            return NEGATIVE_ANSWER;
+        }
+        if (number == array[midIndex]) {
+            return midIndex;
+        } else if (number < array[midIndex]) {
```
hsp-harut: Arman jan, please be attentive. We have discussed this issue 100 times. Do you need "Else " block whenever you returned something in the above "if" statement???

src/main/java/am/hsp/apps/algorithms/search/InterpolationSearch.java
```diff
@@ -0,0 +1,101 @@
...
+        System.out.println(
+                "Recursion Result: " + getResultByInterpolationSearchRecursion(temporaryArray, COMPARABLE_VALUE, 0, 9));
+    }
+
+    @Override
+    public int search(int[] array, int number) {
+        int startIndex = 0;
+        int endIndex = array.length - 1;
+
+        while (isBelongToRange(number, array[startIndex], array[endIndex])) {
```
hsp-harut: I don't know what you here... As we discussed during our lesson, Interpolation search is the same as binary search. The only difference is the logic for choosing pivot index - for binary search it is the middle one, for interpolation search it uses some algorithm for it. So I see no reason why you have here another solution rather than in binary search...

src/main/java/am/hsp/apps/algorithms/search/InterpolationSearch.java
```diff
@@ -0,0 +1,101 @@
...
+            int requiredIndex = getIndexForInterpolationSearch(array, number, startIndex, endIndex);
+
+            if (array[requiredIndex] == number) {
```
hsp-harut: yes - this is what I tried to explain you in the binary search task. This is a more preferable approach, rather than the way you used in binary search task...

src/main/java/am/hsp/apps/algorithms/search/InterpolationSearch.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param array
+     * @param number
+     * @param startIndex
+     * @param endIndex
+     * @return
+     */
+    private int getResultByInterpolationSearchRecursion(int[] array, int number, int startIndex, int endIndex) {
+        if (array[startIndex] == number) {
+            return startIndex;
+        } else {
```
hsp-harut: vay vay vay... es arden mtacum em es urish marda grel...

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,89 @@
...
+     * 
+     * @see main.java.am.hsp.apps.algorithms.search.SearchAlgorithms#search(int[],
+     * int)
+     */
+    @Override
+    public int search(int[] array, int number) {
+        int step = getStepSize(array);
+        int size = array.length - 1;
+        int startIndex = 0;
+        int endIndex = step;
```
hsp-harut: I don't know why you need here 2 variables: startIndex and endIndex.
You just need to have "currentIndex" and step. You need to increase currentIndex by the step, while the element in the currentIndex position (arr[currentIndex]) which is less than the number you are searching for.
So after the while loop only 2 cases possible: 
1. current indexed element is equal to number.
2. current indexed element is greater than the number.

For the second case you just need to to step back and implement linear search.

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,89 @@
...
+
+        while (array[endIndex] <= number && endIndex < size) {
+            startIndex = endIndex;
+            endIndex += step;
+            if (endIndex > size - 1) {
+                endIndex = size;
+            }
+        }
+
+        return getLinearSerch(array, number, startIndex, endIndex);
```
hsp-harut: no. So you are always calling linear search to find an element? no... Based on algorithm logic linear search is required only when you current indexed element is greater than the number you are looking for. So your code should be implemented appropriately. 
You never even checked equals case for the current indexed element and the number... why???

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,89 @@
...
+     * int)
+     */
+    @Override
+    public int search(int[] array, int number) {
+        int step = getStepSize(array);
+        int size = array.length - 1;
+        int startIndex = 0;
+        int endIndex = step;
+
+        while (array[endIndex] <= number && endIndex < size) {
```
hsp-harut: no. there is no need to continue while if you already found the required element (array[endIndex] <**=** number)

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,89 @@
...
+            startIndex = endIndex;
+            endIndex += step;
+            if (endIndex > size - 1) {
```
hsp-harut: extract this into separate method for finding minimum from 2 numbers.

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param array
+     * @param number
+     * @param startIndex
+     * @param endIndex
+     * @return
+     */
+    private int getResultByJumpSearchRecursive(int[] array, int number, int startIndex, int endIndex) {
+        int step = getStepSize(array);
+
+        while (array[endIndex] <= number && endIndex < array.length - 1) {
```
hsp-harut: same... I see no reason t continue with the loop if you found the number...

src/main/java/am/hsp/apps/algorithms/search/InterpolationSearch.java
hsp-arman: Harut jan esi iskakanic urish mard a grel , yes  porcel em es sax maqrel kargi berel .

src/main/java/am/hsp/apps/algorithms/search/InterpolationSearch.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Return result by interpolation search , with recursion method
+     * 
+     * @param array
+     * @param number
+     * @param startIndex
+     * @param endIndex
+     * @return
+     */
+    private int searchByInterpolationRecursively(int[] array, int number, int startIndex, int endIndex) {
```
hsp-harut: same

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,90 @@
...
+
+        if (array[currentIndex] == number) {
+            return currentIndex;
+        }
+
+        while (array[currentIndex] < number) {
+            currentIndex += step;
+        }
+
+        return getLinearSerch(array, number, currentIndex - step, currentIndex);
```
hsp-harut: you didn't fix my comment... Your code will now throw an ArrayIndexOutOfBoundsException if you will test it for the number 168, and will set the step = 7

PR #151, opened 2018-09-27
Hripsime: Sorting Algorithms

hsp-hripsime:


src/main/java/am/hsp/apps/algorithms/search/BinarySearch.java
```diff
@@ -39,7 +39,7 @@ public void solve() {
      * int)
      */
     protected int search(int[] array, int number) {
-        if (isNumberOutOfSortedArrayElementsRange(array, number)) {
+        if (isNumberOutOfSortedArrayElementsRange(array, number, 0, array.length - 1)) {
```
hsp-harut: hmm... I guess there was no need for this change

src/main/java/am/hsp/apps/algorithms/search/BinarySearch.java
```diff
@@ -73,7 +73,7 @@ protected int search(int[] array, int number) {
      * @return
      */
     private int searchRecursively(int[] array, int number, int startIndex, int endIndex) {
-        if (isNumberOutOfSortedArrayElementsRange(array, number)) {
+        if (isNumberOutOfSortedArrayElementsRange(array, number, 0, array.length - 1)) {
```
hsp-harut: same

src/main/java/am/hsp/apps/algorithms/general/GeneralMethods.java
```diff
@@ -110,8 +112,35 @@ protected int getMin(int number1, int number2) {
     }
 
     /**
-     * Checks whether the given number is less than the given sorted array's
-     * first element and more than last element.
+     * <<<<<<< HEAD Checks whether the given number is less than the given sorted
```
hsp-harut: ->

src/main/java/am/hsp/apps/algorithms/search/InterpolationSearch.java
```diff
@@ -41,14 +41,14 @@ public void solve() {
      * @return
      */
     protected int search(int[] array, int number) {
-        if (isNumberOutOfSortedArrayElementsRange(array, number)) {
```
hsp-harut: Please make changes according to our discussion

src/main/java/am/hsp/apps/algorithms/sort/MergeSort.java
```diff
@@ -0,0 +1,170 @@
...
+import main.java.am.hsp.apps.algorithms.sort.base.SortAlgorithms;
+
+/**
+ * Contains methods for sorting the given array in ascending and descending
+ * orders by quick sorting algorithm.
+ * 
+ * @author user
+ *
+ */
+public class MergeSort extends SortAlgorithms {
```
hsp-harut: Leave this class, just rename it MyMergeSort

src/main/java/am/hsp/apps/algorithms/sort/QuickSort.java
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * Sorts and merges the given array in ascending order.
+     * 
+     * @param array
+     * @param lowerIndex
+     * @param higherIndex
+     * @return
+     */
+    private void sortAndMergeInAscendingOrder(int[] array, int lowerIndex, int higherIndex) {
```
hsp-harut: just sort (no merge)

src/main/java/am/hsp/apps/algorithms/search/BinarySearch.java
```diff
@@ -28,21 +28,17 @@ public void solve() {
         System.out.println(
                 "The searchable element index in case of iterative implementation is: " + search(array, number));
         System.out.println("The searchable element index in case of recursive implementation is: "
-                + searchRecursively(array, number, 0, array.length - 1));
+                + searchRecursivelyByAlgorithm(array, number, 0, array.length - 1));
```
hsp-harut: I guess here you need to call "searchRecursively" method.

src/main/java/am/hsp/apps/algorithms/search/base/BinaryInterpolationTemplate.java
```diff
@@ -0,0 +1,54 @@
+package main.java.am.hsp.apps.algorithms.search.base;
+
+public abstract class BinaryInterpolationTemplate extends SearchAlgorithms {
```
hsp-harut: hmm... lets discuss this in our lesson...

src/main/java/am/hsp/apps/algorithms/sort/MergeSort.java
```diff
@@ -0,0 +1,173 @@
...
+     * 
+     * @param array
+     * @param lowerIndex
+     * @param middle
+     * @param higherIndex
+     */
+    private void mergeInAscOrder(int[] array, int lowerIndex, int middle, int higherIndex) {
+        int[] tempArray = new int[array.length];
+
+        for (int i = lowerIndex; i <= higherIndex; i++) {
```
hsp-harut: must be extracted to separate method...

src/main/java/am/hsp/apps/algorithms/search/base/SearchAlgorithmsForSortedArray.java
```diff
@@ -0,0 +1,52 @@
+package main.java.am.hsp.apps.algorithms.search.base;
+
+/**
+ * Contains methods general for searching algorithms in sorted arrays.
+ * 
+ * @author hripsime
+ *
+ */
+public abstract class SearchAlgorithmsForSortedArray extends SearchAlgorithms {
+    protected int startIndex = 0;
```
hsp-harut: no no... Remember - each instance variable should have its strong reason and meaning. 
For example - it is clear, that each searching algorithm should provide a search in the array/collection of the elements, so that is why we have created appropriate intance variable "int[] array". Also it should be the number, which we will try to find by using an algorithm - so that is why we have an appropriate instance variable for it. 

So these 2 instance variables that you have created, have only one reason - to make our job 'easier' (to be honest for me it makes a code more complicated than simplifies it). So you are initializing them first as first and last indexes of the array, and then you are using them as "local variables" - so they have no "state" meaning for the class - they are not defining the state of the "SearchAlgorithmsForSortedArray"-s.
Besides that, based on clean code principles (I am not sure whether you learned that part from the video tutorials of clean code lessons you learned so far, or that comes in further lessons), you must Never create an instance variables where the data is possible to get from already available instance variables. In your case, it is pretty simple to get the last index of the array just by using array.length function...
Anyway, I think you will have questions on this, so lets discuss this in our lesson.

src/main/java/am/hsp/apps/algorithms/sort/MergeSort.java
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * Merges the given two sorted arrays in ascending orders.
+     * 
+     * @param array
+     * @param lowerIndex
+     * @param middle
+     * @param higherIndex
+     */
+    private void mergeInAscOrder(int[] array, int lowerIndex, int middle, int higherIndex) {
+        int[] tempArray = getArrayCopy(array);
```
hsp-harut: Based on algorithm, you need to get a copy only for specified range, and not copy the whole array. 
(If you would need to copy the whole array you could use Arrays.copyOf method rather than implement it manually)

src/main/java/am/hsp/apps/algorithms/sort/MergeSort.java
```diff
@@ -0,0 +1,185 @@
...
+     * 
+     * @param array
+     * @param lowerIndex
+     * @param middle
+     * @param higherIndex
+     */
+    private void mergeInDesOrder(int[] array, int lowerIndex, int middle, int higherIndex) {
+        int[] tempArray = new int[array.length];
+
+        for (int i = lowerIndex; i <= higherIndex; i++) {
```
hsp-harut: be attentive!

src/main/java/am/hsp/apps/algorithms/search/base/BinaryInterpolationSearchTemplate.java
```diff
@@ -0,0 +1,46 @@
...
+        if (startIndex <= endIndex) {
+            int pivotIndex = getPivotIndex();
+
+            if (number == array[pivotIndex]) {
+                return pivotIndex;
+            }
+
+            if (number < array[pivotIndex]) {
+                endIndex = pivotIndex - 1;
+                return search();
```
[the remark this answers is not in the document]
hsp-hripsime: In case of Interpolation search, in each iteration before getting pivot index, at first we must check whether the number is in the current range of the array or not, otherwise we will get a negative index, as the interpolation pivot index formula is working only for the numbers among the range.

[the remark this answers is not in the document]
hsp-harut: If I am not mistaken the issue you have mentioned exists for the algorithm **without Recursion**. Doesn't it?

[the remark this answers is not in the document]
hsp-hripsime: Yes, we hadn't that issue in case of recursive algorithm, as each time calling the function it starts checking whether the number belongs to the range at first. But in our case, we removed that checking part into the search() function, therefore if I return searchByAlgorithm() instead of search(), I will miss that checking part․ So, to include that part I have to return search() function, otherwise will get an infinite loop.

src/main/java/am/hsp/apps/algorithms/search/base/SearchAlgorithmsForSortedArray.java
hsp-hripsime: Honestly, I also don't like the thought of making them as instance variables. The only reason was that I could not find another reasonable solution for it. I need to initialize the first and last indexes dynamically. One solution is to make them parameters of the search() function, but in case of linear and jump algorithms there is no need of such parameters. Then how I can do that?

src/main/java/am/hsp/apps/algorithms/search/base/BinaryInterpolationSearchTemplate.java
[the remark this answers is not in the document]
hsp-harut: ok!

src/main/java/am/hsp/apps/algorithms/search/BinarySearch.java
```diff
@@ -22,58 +22,20 @@ public static void main(String[] args) {
...
-     * the given number returns NOT_FOUND.
+     * Searches the give number by binary search algorithm recursively.
      * 
-     * @param array
-     * @param number
      * @return
      */
-    private int searchRecursively(int[] array, int number, int startIndex, int endIndex) {
-        if (isNumberOutOfSortedArrayElementsRange(array, number)) {
+    protected int searchRecursively(int startIndex, int endIndex) {
```
hsp-harut: This is the same as in interpolation search. So why you didn't make this method a template method by moving it into BinaryInterpolationSearchTemplate class (the same way you did for iterative search)?

PR #153, opened 2018-10-02
Arman: Sorting Algorithms 

hsp-arman:
bubble sort 
selection sort
insertion sort
quick sort 
merge sort


src/main/java/am/hsp/apps/algorithms/sorting/MergeSort.java
```diff
@@ -0,0 +1,87 @@
...
+     * 
+     * @param array
+     * @param low
+     * @param middle
+     * @param high
+     */
+    private void merge(int[] array, int low, int middle, int high) {
+        int[] tempArray = new int[array.length];
+
+        for (int i = low; i <= high; i++) {
```
hsp-harut: Don't you see that this is completely separate functionality which creates a copy of given array of given range? COHESION

src/main/java/am/hsp/apps/algorithms/sorting/QuickSort.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * The recursive method makes the same operation for all the pieces
+     * 
+     * @param array
+     * @param startIndex
+     * @param endIndex
+     */
+    private void sortByRecursion(int array[], int startIndex, int endIndex) {
+        int currentIndex = quickSort(array, startIndex, endIndex);
```
hsp-harut: I don't even understand what you did here... Your method name is "quickSort", but you are using it to get some index?

Please have a look on the example I gave you as a starting point, and update your current code appropriately...
http://www.java2novice.com/java-sorting-algorithms/quick-sort/

src/main/java/am/hsp/apps/algorithms/sorting/SelectionSort.java
```diff
@@ -0,0 +1,48 @@
...
+     * (non-Javadoc)
+     * 
+     * @see main.java.am.hsp.apps.algorithms.sorting.SortAlgorithms#sort(int[])
+     */
+    @Override
+    public int[] sort(int[] array) {
+
+        for (int i = 0; i < array.length - 1; i++) {
+            for (int j = i + 1; j < array.length; j++) {
+                if (array[i] > array[j]) {
```
hsp-harut: No... You didn't get the idea of an algorithm... It is not swaping elements each time it finds an element which is less than "i" element... It must find MINIMUM (or max) element from the array starting from "i", and swap it with the "i" element...

src/main/java/am/hsp/apps/algorithms/sorting/SortAlgorithms.java
```diff
@@ -0,0 +1,29 @@
...
+import main.java.am.hsp.apps.seua.base.TaskArray;
+
+/**
+ * Contains all general methods necessary for sorting tasks.
+ * 
+ * @author arman
+ *
+ */
+public abstract class SortAlgorithms extends TaskArray {
+    /*
```
hsp-harut: Wrong comment type for the method...

src/main/java/am/hsp/apps/algorithms/sorting/MergeSort.java
```diff
@@ -0,0 +1,112 @@
...
+        return (low + high) / 2;
+    }
+
+    /**
+     * Return similar of given array
+     * 
+     * @param array
+     * @return
+     */
+    private int[] getSimilarArray(int[] array) {
```
hsp-harut: Wrong. The original functionality was not copying the whole array. It just copied only the given range of the array...

image

src/main/java/am/hsp/apps/algorithms/sorting/SelectionSort.java
```diff
@@ -0,0 +1,52 @@
...
+     * @see main.java.am.hsp.apps.algorithms.sorting.SortAlgorithms#sort(int[])
+     */
+    @Override
+    public int[] sort(int[] array) {
+        int length = array.length;
+
+        for (int i = 0; i < length; i++) {
+            int min = i;
+            
+            for (int j = i + 1; j < length; j++) {
```
hsp-harut: Don't you see that this is a functionality for finding a minimum from the array starting from given index? So Why You Didn't Extract It To Separate Method?!!!!!

src/main/java/am/hsp/apps/algorithms/sorting/SortAlgorithms.java
hsp-harut: You didn't fix my comment!!!!!

Comment type for the method must be: '/** */' and not '/* */'

src/main/java/am/hsp/apps/algorithms/sorting/SelectionSort.java
```diff
@@ -0,0 +1,68 @@
...
+        return array;
+    }
+   
+    /**
+     * Return index of minimum element , from the given index to the end 
+     * @param array
+     * @param index
+     * @return
+     */
+    private int getMin(int[] array, int index) {
```
hsp-harut: "startIndex" would be more descriptive.

src/main/java/am/hsp/apps/algorithms/sorting/SelectionSort.java
```diff
@@ -0,0 +1,68 @@
...
+        int length = array.length;
+        int min = array[index];
+        int minIndex = index;
+        
+        for (int i = index + 1; i < length; i++) {
```
hsp-harut: It is better to start with the startIndex and not from "startIndex + 1". You can use this method by passing "i + 1" as a param, rather than have "startIndex + 1" in the for loop

PR #165, opened 2018-11-21
Gor: tasks 51 60

hsp-gor:
Implemented tasks 51-60


src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -164,4 +164,20 @@ protected boolean isGeometricallProgrssion(int number1, int number2, int number3
...
+	/**
+	 * Method returns power of the number
+	 * 
+	 * @param number
+	 * @param degree
+	 * @return
+	 */
+	protected int pow(int number, int degree) {
+
+		int temp = 1;
```
hsp-harut: Bad variable name. 
This is actually the RESULT - so the variable name must be **result**...

src/main/java/am/hsp/apps/seua/base/TaskDigits.java
```diff
@@ -0,0 +1,137 @@
+package main.java.am.hsp.apps.seua.base;
+
+public abstract class TaskDigits extends TaskBase {
```
hsp-harut: COMMENT

src/main/java/am/hsp/apps/seua/base/TaskDigits.java
```diff
@@ -0,0 +1,137 @@
...
+	}
+	
+	/**
+	 * Method return division of numbers
+	 * 
+	 * @param number1
+	 * @param number2
+	 * @return
+	 */
+	protected double getDivision(double number1, double number2) {
```
hsp-harut: As I said no need to have methods for already defined operations...

src/main/java/am/hsp/apps/seua/base/TaskDigits.java
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	/**
+	 * Method returns unit digit of the number
+	 * 
+	 * @param number
+	 * @return
+	 */
+	protected int getUnitDigit(int number) {
+		return getDigitsByIndex(number, UNIT_DIGIT);
```
hsp-harut: No. Unit digit is actually the last digit of the number, which could be extracted just by using **number % 10**
So why are you using another method, where you are doing unneeded other operations such as **number = number / pow(10, index - 1);**

src/main/java/am/hsp/apps/seua/base/TaskDigits.java
```diff
@@ -0,0 +1,137 @@
...
+	 * Method returns digit by index
+	 * 
+	 * @param number
+	 * @param index
+	 * @return
+	 */
+	protected int getDigitsByIndex(int number, int index) {
+
+		number = number / pow(10, index - 1);
+		return number % 10;
```
hsp-harut: DRY violation. You have this code written in many places - but this is actually last digit of the number, or the units of the number - so you must use the appropriate method here (and EVERYWHERE IN YOUR CODE)...

src/main/java/am/hsp/apps/seua/base/TaskDigits.java
```diff
@@ -0,0 +1,137 @@
...
+	 */
+	protected int getSumOfDigits(int number) {
+		
+		int result = 0;
+		while (number > 0) {
+			
+			int unitdigit = getUnitDigit(number);	
+			result += unitdigit;
+			
+			number /= 10;
```
hsp-harut: You are repeating this code in many methods - so you are violating DRY principle.
You need to create a method which will remove the last digit from the number and will return the result...

src/main/java/am/hsp/apps/seua/base/TaskDigits.java
```diff
@@ -0,0 +1,137 @@
...
+	 * 
+	 * @param number
+	 * @return
+	 */
+	protected int getSumOfDigits(int number) {
+		
+		int result = 0;
+		while (number > 0) {
+			
+			int unitdigit = getUnitDigit(number);
```
hsp-harut: Wrong naming format

src/main/java/am/hsp/apps/seua/base/TaskDigits.java
```diff
@@ -0,0 +1,137 @@
...
+	 */
+	protected int getMaxOfDigits(int number) {
+		
+		int max = getUnitDigit(number);
+		
+		while(number > 0) {
+			
+			int unitdigit = getUnitDigit(number);
+			
+			if(max < unitdigit) {
```
hsp-harut: Again. We discussed this in our lesson... When you are trying to find a maximum, you need to check whether current number is GREATER than the maximum, and if it is - then you need to store it as a maximum.
So maximum means something which is GREATER than all others - so in your code you need to find a number which is greater than the maximum!
So having said this, you need to reflect it in the "if" condition - so you need to check whether current digit is Greater than the maximum , and if yes - you need to update the maximum value by the current digit.

src/main/java/am/hsp/apps/seua/base/TaskDigits.java
```diff
@@ -0,0 +1,137 @@
...
+	 */
+	protected int getMinOfDigits(int number) {
+		
+		int min = getUnitDigit(number);
+		
+		while(number > 0) {
+			
+			int unitdigit = getUnitDigit(number);
+			
+			if(min > unitdigit) {
```
hsp-harut: same

src/main/java/am/hsp/apps/seua/base/TaskDigits.java
```diff
@@ -0,0 +1,137 @@
+package main.java.am.hsp.apps.seua.base;
+
+public abstract class TaskDigits extends TaskBase {
+
+	private final int UNIT_DIGIT = 1;
```
hsp-harut: You can use here "protected" to be able to use this constants also in the Tasks which extended from TaskDigit class

src/main/java/am/hsp/apps/seua/tasks_51_60/Task_52.java
```diff
@@ -0,0 +1,35 @@
...
+		System.out.println(isAnyDigitsEqual(number));
+	}
+	
+	/**
+	 * Method return true if there is any equal digits 
+	 * @param number
+	 * @return
+	 */
+	private boolean isAnyDigitsEqual(int number) {
+		return isEqualDigitsByIndex(number, 1, 2) || isEqualDigitsByIndex(number, 1, 3) || isEqualDigitsByIndex(number, 2, 3);
```
hsp-harut: magic numbers

src/main/java/am/hsp/apps/seua/tasks_51_60/Task_53.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	/*
+	 * (non-Javadoc)
+	 * 
+	 * @see main.java.am.hsp.apps.seua.base.TaskBase#solve()
+	 */
+	public void solve() {
+		int number = getUserInput("Please enter the number");
+		int tasknumber = getUserInput("Please enter the number");
```
hsp-harut: 1. WRONG NAMING FORMAT
2. You are displaying the SAME message to user saying to enter the number - it is not clear what is the purpose of each number that you are asking to input. So your messages should be descriptive, so the user will understand what to input...

src/main/java/am/hsp/apps/seua/tasks_51_60/Task_56.java
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	/*
+	 * (non-Javadoc)
+	 * 
+	 * @see main.java.am.hsp.apps.seua.base.TaskBase#solve()
+	 */
+	public void solve() {
+		int number = getUserInput("Please enter the number");
+		System.out.println(getUnitDigit(number) > getTensDigit(number) ? getDivisionSumOfDigitsByNumber(number) : number);
```
hsp-harut: This makes a "solve" method to be Non Cohesive. "solve" method must not contain any task specific logic - it should be only responsible for the communication with the user.

So you need to create a new method lets say "getResult" which will return the required result based on task specific logic.

You need to use the same approach for all other tasks as well...

src/main/java/am/hsp/apps/seua/tasks_51_60/Task_57.java
```diff
@@ -0,0 +1,44 @@
...
+			System.out.println(number > SUCCES_CASE ? getDivision(getTensDigit(number), getUnitDigit(number)) : getDivision(getHundredsDigit(number), getUnitDigit(number)));
+		}
+	}
+
+	/**
+	 * method returns false if there is digit equal 0 and true if non of digits are 0 
+	 * @param number
+	 * @return
+	 */
+	private boolean isAnyDigitEqualNull(int number) {
```
hsp-harut: 1. Your method name does not correspond to its implementation. Based on your method name it should return "true" when there is ANY digit equal to "0". But based on your method implementation it returns "true" when ALL the digits are NOT EQUAL to "0"

2. Why do you need this method at all? If you tried to avoid a situation when there will be a division to "0", then why are you checking this for all the digits? Based on your task it is only needed to divide into UNIT digit, so it is enough to check
that the unit digit IS NOT NULL

src/main/java/am/hsp/apps/seua/tasks_51_60/Task_58.java
```diff
@@ -0,0 +1,36 @@
...
+	}
+	
+	/*
+	 * (non-Javadoc)
+	 * @see main.java.am.hsp.apps.seua.base.TaskBase#solve()
+	 */
+	public void solve() {
+		int number = getUserInput("Please enter the number");
+		char f = 'a';
+		System.out.println(sumOfHundredsTensDigits(number) > SUCCES_CASE ? f = 'b' : f );
```
hsp-harut: image

src/main/java/am/hsp/apps/seua/tasks_51_60/Task_57.java
```diff
@@ -0,0 +1,37 @@
...
+	 * 
+	 * @see main.java.am.hsp.apps.seua.base.TaskBase#solve()
+	 */
+	public void solve() {
+		
+		int number = getUserInput("Please enter the number");
+		
+		if(getUnitDigit(number) != 0) {
+		
+			System.out.println(number > SUCCES_CASE ? getTensDigit(number) / getUnitDigit(number)
```
hsp-harut: You didn't fix my comment:

image

src/main/java/am/hsp/apps/seua/base/TaskBase.java
hsp-harut: YOU DIDN't FIX MY COMMENT

src/main/java/am/hsp/apps/seua/tasks_51_60/Task_57.java
```diff
@@ -0,0 +1,37 @@
...
+	/*
+	 * (non-Javadoc)
+	 * 
+	 * @see main.java.am.hsp.apps.seua.base.TaskBase#solve()
+	 */
+	public void solve() {
+		
+		int number = getUserInput("Please enter the number");
+		
+		if(getUnitDigit(number) != 0) {
```
hsp-harut: See my comment below... Solve METHOD MUST NOT contain any task specific logic - that makes it to be non cohesive

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -164,4 +164,29 @@ protected boolean isGeometricallProgrssion(int number1, int number2, int number3
...
+		}
+		return result;
+	}
+	
+	/**
+	 * method reterns square of the number
+	 * @param number
+	 * @return
+	 */
+	protected int square(int number) {
```
hsp-harut: Why do you need this method when you have a pow method?

src/main/java/am/hsp/apps/seua/tasks_51_60/Task_57.java
```diff
@@ -0,0 +1,48 @@
...
+	 * @param number
+	 * @return
+	 */
+	private double getDivisionOfDigits(int number) {
+		return number > SUCCES_CASE ? getTensDigit(number) / getUnitDigit(number)
+				: getHundredsDigit(number) / getUnitDigit(number);
+
+	}
+	
+	private boolean isValiedNumber(int number) {
```
hsp-harut: Es inch es gre Gor jan?????

src/main/java/am/hsp/apps/seua/tasks_51_60/Task_58.java
```diff
@@ -0,0 +1,45 @@
...
+	}
+	
+	/**
+	 * method returns 'b' if sum of digits is greater than the given number
+	 * @param number
+	 * @return
+	 */
+	private char getResult(int number) {
+		char f = 'a';
+		return sumOfHundredsTensDigits(number) > SUCCES_CASE ? f = 'b' : f ;
```
hsp-harut: No... should be:
char f = sumOfHundredsTensDigits(number) > SUCCES_CASE ? 'b' : 'a' ;
return f;

PR #181, opened 2018-12-10
Shant tasks 161 170

hsp-shant:
Please check out the implemented files for further checks.


src/main/java/am/hsp/apps/seua/task_161_170/Task_167.java
```diff
@@ -0,0 +1,53 @@
...
+
+	}
+
+	/**
+	 * Print true if there is any sin X with n degree got minus correct number,
+	 * where n degree are numbers range from number 1 to number 30 else print false.
+	 * 
+	 * @return
+	 */
+	private boolean isAnySinXGotMinuesNDegreeNumber(int number) {
```
hsp-harut: Non cohesive method. It does 2 things:
a) checks whether is the required condition true for any number in the given range.
b) Calculates the sinus for the number

src/main/java/am/hsp/apps/seua/task_161_170/Task_168.java
```diff
@@ -0,0 +1,40 @@
...
+
+	}
+
+	/**
+	 * Print true if the input number is a prime number else print false
+	 * 
+	 * @param number
+	 * @return
+	 */
+	private boolean isNumberPrime(int number) {
```
hsp-harut: WHY DO YOU NEED THIS METHOD ???

src/main/java/am/hsp/apps/seua/task_161_170/Task_170.java
```diff
@@ -0,0 +1,49 @@
...
+		int number = getUserInput("Please enter your number");
+		System.out.println(getGreaterSquareRootNumberOfInputNumber(number));
+
+	}
+
+	/**
+	 * Print the number which is greater than the input number and which is square
+	 * root correct number
+	 */
+	private int getGreaterSquareRootNumberOfInputNumber(int number) {
```
hsp-harut: Wrong implementation. 
The task says to find the number which is greater than the given number and which is equal to value of 2 raised to the power of any integer: 
yourNumber = pow(2, n)

This means for example if your user input = 8, then your program must return 16 (which is actually pow(2, 4))
Currently your program returns 9 for the sample above...

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -101,31 +115,147 @@ protected boolean isMathProgression(int number1, int number2, int number3) {
...
+	 */
+	protected boolean isNotMultiplier(int multiplier, int number) {
+		return !isMultiplier(multiplier, number);
+
+	}
+
+	/**
+	 * Check if is division by balance
+	 */
+	protected boolean isDivisionByBalance(int number, int divisor, double balance) {
```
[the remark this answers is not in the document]
hsp-shant: I just thinking that i will have double in result, like 9/2 = 4.5,  this is why.

src/main/java/am/hsp/apps/seua/task_161_170/Task_170.java
```diff
@@ -0,0 +1,49 @@
...
+		int number = getUserInput("Please enter your number");
+		System.out.println(getGreaterSquareRootNumberOfInputNumber(number));
+
+	}
+
+	/**
+	 * Find the number which is greater than the given number and which is equal to value of 2 raised to the power of any integer
+	 */
+	private int getGreaterSquareRootNumberOfInputNumber(int number) {
+		number++;
```
hsp-harut: Why you have incremented the number? Can't you just have a correct condition in the "while" loop?

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -139,50 +185,76 @@ protected boolean isGeoProgresion(int number1, int number2, int number3) {
...
 
-		while (result < number) {
-			result = pow(i, 2);
-			i++;
+	}
 
+	/**
+	 * Return true if the given number is exponent of base, otherwise return false.
+	 */
+	protected boolean isNumberExponentOfBase(int number, int base) {
```
hsp-harut: YOur method will not work correctly. Test your code before pushing changes for review.

src/main/java/am/hsp/apps/seua/task_161_170/Task_167.java
```diff
@@ -0,0 +1,51 @@
...
+
+		for (int i = START_RANGE; i <= END_RANGE; i++) {
+			if (Math.sin(pow(number, i)) < 0.0) {
```
hsp-harut: NOT COHESIVE METHOD.
Your method is doing 2 things:
a) Checks the required condition for the range of numbers and return appropriate boolean
b) The actual condition implementation is also here

YOu need to have a separate method which will check whether the sinus for the given number is negative or not.
Then you can use that method here, by passing as an argument the number raised to the power of "i"

src/main/java/am/hsp/apps/seua/task_161_170/Task_170.java
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * Find the number which is greater than the given number and which is equal to value of 2 raised to the power of any integer
+	 */
+	private int getGreaterSquareRootNumberOfInputNumber(int number) {
+		
+		int result;
+		int i = 1;
+		do {
+			result = pow(BASE, ++i);
+		} while (number++ > result);
```
hsp-harut: ???? What you are incrementing the number after each iteration??????

Shant if you don't understand my comments then you MUST ASK QUESTIONS and not just doing I don't know what and give it to review.

This was your previous version of this method:
image

And I just meant in my comment that there is no need to increment the number at all. All you needed there is just to have a correct condition in the while loop: 
**while (result <= number)**

AND THAT'S ALL

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -139,50 +185,75 @@ protected boolean isGeoProgresion(int number1, int number2, int number3) {
...
+		
+		while (number > 1) {
```
hsp-harut: You didn't fix my comment. Your method can not work correctly? Why don't you testing I don't understand???
Is the "7" any exponent of "2" ??? Of course no. But your code returns TRUE!

I can not understand why are you removing needs work label when you didn't fix anything???

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -139,50 +185,79 @@ protected boolean isGeoProgresion(int number1, int number2, int number3) {
...
+	}
+
+	/**
+	 * Return true if the given number is exponent of base, otherwise return false.
+	 */
+	protected boolean isNumberExponentOfBase(double number, int base) {
+		
+		while (number > 1) {
+			number /= base;
+			if (number == 1) {
```
hsp-harut: No need for this "if" block at all. All you need is to have a correct "return" by the end of the method: 
return number == 1;

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -139,50 +185,79 @@ protected boolean isGeoProgresion(int number1, int number2, int number3) {
...
+				return true;
+			}
 
 		}
-		return result == number;
+		return false;
```
hsp-harut: see my comment above

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -139,50 +185,79 @@ protected boolean isGeoProgresion(int number1, int number2, int number3) {
...
-		while (result < number) {
-			result = pow(i, 2);
-			i++;
+	}
+
+	/**
+	 * Return true if the given number is exponent of base, otherwise return false.
+	 */
+	protected boolean isNumberExponentOfBase(double number, int base) {
+
```
hsp-harut: you need to check here the case when number is less than the "base", and return false for that case. Otherwise your method will not work correctly for the cases when "number == 1".

src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -2,12 +2,15 @@
 
 import java.util.Scanner;
 
+import javafx.scene.control.SplitPane.Divider;
```
hsp-harut: Wrong import which causes for me compilation errors.

PR #200, opened 2019-01-14
Vahe:  tasks 221 230

hsp-vahe:


src/main/java/am/hsp/apps/seua/base/TaskBase.java
```diff
@@ -235,6 +235,19 @@ protected int getAbsoluteValueOfGivenNumber(int number) {
...
+	 * Is number belong to the range.
+	 */
+	protected boolean isBelongInnerRange(int number, int startRange, int endRange) {
+			return number > startRange && number < endRange;
+	}
+	
+	/**
+	 * Is subtraction of two numbers positive.
+	 */
+	protected boolean isPositive(int firstNumber, int secondNumber) {
```
hsp-harut: Based on method name it is not clear what should be positive? Maybe "difference"?

src/main/java/am/hsp/apps/seua/tasks_221_230/Task_229.java
```diff
@@ -0,0 +1,54 @@
...
+		System.out.println("product of the elements which`s value and index subtraction is positive number_ " + getResult(array));
+	}
+	
+	/**
+	 * Get the product of the elements which`s value and index subtraction is positive number.
+	 */
+	private int getResult(int array[]) {
+		int product = 1;
+		for(int i = 0; i < array.length; i++) {
+			if(isPositive(array[i], i)) {
```
hsp-harut: I guess that would be more correct to have a method which will accept an array as an argument, and the index of the element for which you need to check whether the difference of the value and the index is positive or not. That would be a meaningful method definitely. 
You see, it is not a useful method to check whether the difference between two given parameters is positive or not. It is hard to imagine where we might need to have such functionality.
**isDifferenceOfValueAndIndexPositive**

PR #215, opened 2019-02-06
Grigor: Search_Aalgorithms

hsp-grigor:


src/main/java/am/hsp/apps/algorithms/Binary_Search.java
```diff
@@ -0,0 +1,66 @@
...
+
+	}
+
+	/*
+	 * (non-Javadoc)
+	 * 
+	 * @see main.java.am.hsp.apps.seua.base.TaskBase#solve()
+	 */
+	@Override
+	public void solve() {
```
hsp-harut: Searching algorithms are not "solving" anything. They are just doing "search"

src/main/java/am/hsp/apps/algorithms/Binary_Search.java
```diff
@@ -0,0 +1,66 @@
...
+	/**
+	 * Find input number index in array using binary search algorithms. 
+	 * @param array
+	 * @param start
+	 * @param end
+	 * @param searchNumber
+	 * @return
+	 */
+	private int searchNumberIndex(int[] array, int start, int end, int searchNumber) {
+		int pivot = getPivot(start, end);
```
hsp-harut: You need to get the pivot only when it is actual. So it is not needed when you already know that the searchNumber does not exist in your array...

src/main/java/am/hsp/apps/algorithms/Binary_Search.java
```diff
@@ -0,0 +1,66 @@
...
+		
+		if (start > end || array[start] > searchNumber || array[end] < searchNumber) {
```
hsp-harut: This makes your code to be non cohesive. You need to create a method which will check whether searchNumber does not exist.
**isNotExist**

src/main/java/am/hsp/apps/algorithms/Binary_Search.java
```diff
@@ -0,0 +1,66 @@
...
+
+import main.java.am.hsp.apps.seua.base.TaskArray;
+
+/**
+ * Find input number index in array using binary search algorithms
+ * @author GRIGOR
+ *
+ */
+public class Binary_Search extends TaskArray {
+	private static final int NOT_FOUNDE = 0;
```
hsp-harut: 1. NOT_FOUND
2. Wrong value for the constant. 0 is the first index of the array, so when your program shows "0" as a searching result, it is not clear whether it found the search number in the first index of the array or the number is not found at all...

src/main/java/am/hsp/apps/algorithms/Interpolation_Search.java
```diff
@@ -0,0 +1,67 @@
...
+
+import main.java.am.hsp.apps.seua.base.TaskArray;
+
+/**
+ * Find input number index in array using interpolation search algorithms.
+ * 
+ * @author GRIGOR
+ *
+ */
+public class Interpolation_Search extends TaskArray {
```
hsp-harut: same issues in this class

src/main/java/am/hsp/apps/algorithms/Binary_Search.java
```diff
@@ -0,0 +1,66 @@
...
+	/*
+	 * (non-Javadoc)
+	 * 
+	 * @see main.java.am.hsp.apps.seua.base.TaskBase#solve()
+	 */
+	@Override
+	public void solve() {
+		int[] testArray = { 1, 2, 3, 4, 5, 7, 8, 9, 11 };
+		int serachNumber = getUserInput("Please enter the number");
+		System.out.println(searchNumberIndex(testArray, NOT_FOUNDE, testArray.length - 1, serachNumber));
```
hsp-harut: No. It is completely wrong to use NOT_FOUND constant as a method parameter. It does not matter that it has the same value as you need to pass to the method. The purpose of the Constants is to make the code clean and understandable. But in your case, you make the code Very Confusing and not understandable...

src/main/java/am/hsp/apps/algorithms/Binary_Search.java
```diff
@@ -0,0 +1,66 @@
+package main.java.am.hsp.apps.algorithms;
+
+import main.java.am.hsp.apps.seua.base.TaskArray;
+
+/**
+ * Find input number index in array using binary search algorithms
```
hsp-harut: It will also help you in future, to indicate also the complexity of the Algorithm. And also maybe to have the link you used for the algorithm investigation.

src/main/java/am/hsp/apps/algorithms/Jump_Search.java
```diff
@@ -0,0 +1,76 @@
...
+
+	/**
+	 * Find input number index in array using jump search algorithms.
+	 * 
+	 * @param array
+	 * @param searchNumber
+	 * @return
+	 */
+	private int searchNumberIndex(int[] array, int searchNumber) {
+		int start = 0;
```
hsp-harut: Why do you need this variable at all? If just to use in the line 54, then why you are not calculating startIndex for the linearSearch based on your "end" and the "step"?

src/main/java/am/hsp/apps/algorithms/Jump_Search.java
```diff
@@ -0,0 +1,76 @@
...
+		int end = 0;
```
hsp-harut: Actually this is your "current position" in the algorithm. So this variable name could be just **currentIndex**

src/main/java/am/hsp/apps/algorithms/Jump_Search.java
```diff
@@ -0,0 +1,76 @@
...
+		int step = getNearlyRoot(array.length);
+		while (array[end] < searchNumber) {
```
hsp-harut: hmm... yes currently your code works correct from the functional perspective, but actually lets make it also "logically correct".
So algorithm says that you are free to jump forward by the "step" if you didn't reach array last index yet.
So your loop condition should be just: **while (currentIndex < array.length)**

src/main/java/am/hsp/apps/algorithms/Jump_Search.java
```diff
@@ -0,0 +1,76 @@
...
+			start = end;
+			end += step;
```
hsp-harut: Actually, you need jump by the step only when your current element (in currentIndex) is less than the number you are searching for. So it is too soon to make this functionality here.

src/main/java/am/hsp/apps/algorithms/Jump_Search.java
```diff
@@ -0,0 +1,76 @@
...
+	 * @param array
+	 * @param searchNumber
+	 * @return
+	 */
+	private int searchNumberIndex(int[] array, int searchNumber) {
+		int start = 0;
+		int end = 0;
+		int step = getNearlyRoot(array.length);
+		while (array[end] < searchNumber) {
+			start = end;
```
hsp-harut: this is not needed

src/main/java/am/hsp/apps/algorithms/Jump_Search.java
```diff
@@ -0,0 +1,76 @@
...
+			end += step;
+			if (end >= array.length) {
```
hsp-harut: Actually you are doing here again the same "step forward" functionality - so you are initializing the current index by the last array index if the step forward makes you out of array bounds. So as I said this is the "part" of step forward functionality, which should be done only when your current element (in currentIndex) is less than the number you are searching for. So it is too soon to make this functionality here.

src/main/java/am/hsp/apps/algorithms/Jump_Search.java
```diff
@@ -0,0 +1,76 @@
...
+				end = array.length - 1;
+				break;
+			}
+			if (array[end] == searchNumber) {
```
hsp-harut: :+1:

src/main/java/am/hsp/apps/algorithms/Jump_Search.java
```diff
@@ -0,0 +1,76 @@
...
+				return end;
+			}
```
hsp-harut: So after the above "if" block, you already know, that your current element is not equal to the number you are searching for. So you have 2 possible scenarios here:
1. The current element is greater than the number. So you need to make a linearSearch for this case.
So you need to write appropriate "if" block now, and return the "linearSearch"

2. The current element is less than the number. So for this case you just need to make a step forward. So after the "if" block mentioned in "Point 1" above, you just need to reinitialize your **currentIndex** (by making a step forward)

src/main/java/am/hsp/apps/algorithms/Jump_Search.java
```diff
@@ -0,0 +1,76 @@
...
+		}
+		return searchNumberLiner(array, start, end, searchNumber);
```
hsp-harut: Actually, this should be the "NOT_FOUND" case...

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,88 @@
...
+		int step = getNearlyRoot(array.length);
+		while (array[end] <= searchNumber) {
+			end += step;
+			if (end >= array.length) {
+				end = array.length - 1;
+				break;
+			}
+			if (array[end] == searchNumber) {
+				return end;
+			}
```
hsp-harut: see my previous comment:

image

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,88 @@
...
+		}
+		return searchNumberLiner(array, end-step, end, searchNumber);
```
hsp-harut: you didn't fix my comment

image

src/main/java/am/hsp/apps/algorithms/search/base/SearchAlgorithmsBase.java
```diff
@@ -0,0 +1,22 @@
...
+ * @author GRIGOR
+ *
+ */
+public abstract class SearchAlgorithmsBase {
+
+	protected static final int NOT_FOUND = -1;
+
+	protected int[] array;
+
+	public void setArray(int[] array) {
```
hsp-harut: Setter/Getter methods should be the last methods in the class...

src/main/java/am/hsp/apps/algorithms/search/base/SearchAlgorithmsSortedBase.java
```diff
@@ -0,0 +1,25 @@
...
+	/**
+	 * Check whether is that valid to continue the searching process.
+	 * 
+	 * @param array
+	 * @param start
+	 * @param end
+	 * @param searchNumber
+	 * @return
+	 */
+	protected boolean isValidToContinue(int[] array, int start, int end, int searchNumber) {
```
hsp-harut: As I mentioned previously, the "array" is not needed here as a method param...

src/main/java/am/hsp/apps/algorithms/search/DemoSearch.java
```diff
@@ -0,0 +1,76 @@
...
+		demo.demoInterpolationSearch();
+		
+		
+
+	}
+
+	/**
+	 * Demonstration of the binary search algorithm.
+	 */
+	private void demoBinerySearch() {
```
hsp-harut: Binary

src/main/java/am/hsp/apps/algorithms/search/DemoSearch.java
```diff
@@ -0,0 +1,76 @@
...
+	private void demoJumpSearch() {
+		System.out.println("Demonstrating jump search: ");
+		doDemo(new JumpSearch());
+	}
+	
+	/**
+	 * Demonstration of the jump search algorithm.
+	 */
+	private void demoLinerSearch() {
+		System.out.println("Demonstrating jump search: ");
```
hsp-harut: linear

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,89 @@
...
+public class JumpSearch extends SearchAlgorithmsSortedBase {
+
+	/**
+	 * Find input number index in array using jump search algorithms.
+	 * 
+	 * @param array
+	 * @param searchNumber
+	 * @return
+	 */
+	private int searchNumberIndex(int searchNumber) {
```
hsp-harut: What is the reason of this method, if it has the same arguments as the "search" method has?

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,89 @@
...
+		if (isValidToContinue(0, array.length - 1, searchNumber)) {
+			return NOT_FOUND;
+		}
+		int currentIndex = 0;
+		int step = getNearlyRoot(array.length);
+		while (currentIndex < array.length) {
+			if (array[currentIndex] < searchNumber) {
```
hsp-harut: no need for this "if" block here. After 2 "if" blocks below, you can just make a step forward, as the only possible case is when "array[currentIndex] < searchNumber"

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,89 @@
...
+				currentIndex += step;
```
hsp-harut: You are doing Wrong step forward. Test your code when your array consists of 11 elements, and try to find the last element in the array. You will get index out of bounds exception

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,75 @@
...
+				return currentIndex;
+			}
+			if (array[currentIndex] > searchNumber) {
+				return searchNumberLiner(currentIndex - step, currentIndex, searchNumber);
+			}
+
+			currentIndex += step;
+
+		}
+		return searchNumberLiner(currentIndex - step, currentIndex, searchNumber);
```
hsp-harut: No. This is not found case, you have no right to make again linear search here.
And you again change your code to incorrect version **when I 2 times told you that it is Incorrect.** Bad.

image

src/main/java/am/hsp/apps/algorithms/search/JumpSearch.java
```diff
@@ -0,0 +1,75 @@
...
+		int step = getNearlyRoot(array.length);
+		while (currentIndex < array.length) {
+			if (array[currentIndex] == searchNumber) {
+				return currentIndex;
+			}
+			if (array[currentIndex] > searchNumber) {
+				return searchNumberLiner(currentIndex - step, currentIndex, searchNumber);
+			}
+
+			currentIndex += step;
```
hsp-harut: And again... 

image

PR #238, opened 2019-09-29
Tatev tasks 421 450

tatev-hsp:


src/main/java/am/hsp/apps/seua/base/TaskArray.java
```diff
@@ -37,7 +37,24 @@ protected int getRandomNumber(int rangeStart, int rangEnd) {
...
+	 * 
+	 * generate elements from given rang [RANG_START,RANG_END] array
+	 * 
+	 * @param num
+	 * @return
+	 */
+	protected int[][] generateSquareArray(int num) {
+		int array[][] = new int[num][num];
+		for (int i = 0; i < num; i++) {
+			for (int j = 0; j < num; j++) {
```
hsp-harut: hmm... you can optimize your code. Actually the second "for" loop is just for generating the array of "num" elements.
I guess you already have a method which stands for it. So you are violating the DRY principle.
All you need is just to initialize the "i" element of your Square Array (i => each new row) as the newly generated array (second "for" loop - which should be replaced by the appropriate method usage)

src/main/java/am/hsp/apps/seua/tasks_421_450/Task_446.java
```diff
@@ -0,0 +1,54 @@
...
+	 * 
+	 * @param array[][]
+	 * @return
+	 */
+	private int getResult(int[][] array) {
+		int count = 0;
+
+		for (int i = 0; i < array.length; i++) {
+			for (int j = i; j < array.length; j++) {
+				if ((array[i][j] >= START) && (array[i][j] <= END)) {
```
hsp-harut: Hmm... don't you have a method for this? If I am not mistakes you had "isInRange" method difined in one of the base classes. Doesn't it make the deal?

